PR #22, opened 2025-11-07
Refactor/distribution

LeonidElkin:
Done:
 - add discrete distribution support
 - refactore old registry graph
 - add more tests to new graph
 - delete surprisingly useless GenericCharacteristic
 
 Need to be done:
 - Lots of tests and thoroughly checked current
 - Lots of flexible fitters
 - Log likelyhood moved to indefinitvie
 - example update


Desiment:
1. It seems more appropriate to add ``support`` property to Distribution protocol (note that support is mathematical terminology, not support object)
2. I guess current constraint system is not optimal in a sense that a lot of things hardcoded to distribution type. On other the other hand we want keep constraint system non-public. This is the problem we need to solve, you suggested something nera Visitor pattern

src/pysatl_core/distributions/distribution.py
```diff
@@ -61,109 +55,17 @@ def sampling_strategy(self) -> SamplingStrategy: ...
...
-        hence values are computed element-wise.
-        """
-        name = "pdf" if self.distribution_type.kind == "continuous" else "pmf"
-        method = self.computation_strategy.query_method(name, self)
-        xs = np.asarray(batch.array, dtype=np.float64).ravel()
-        vals = np.fromiter((float(method(float(x))) for x in xs), dtype=np.float64, count=xs.size)
-        if np.any(vals <= 0.0):
-            return float("-inf")
-        return float(np.sum(np.log(vals)))
+    def log_likelihood(self, batch: Sample) -> float: ...
```
Desiment: We decided to remove this method?

src/pysatl_core/distributions/fitters.py
```diff
@@ -338,3 +339,270 @@ def f(q: float) -> float:
...
+    support = getattr(distribution, "_support", None)
+    if support is None:
+        raise RuntimeError("Discrete support is required for pmf->cdf.")
+    pmf_func = _resolve(distribution, PMF)
+
+    def _cdf(x: float) -> float:
+        s = 0.0
+        for k in support.iter_leq(x):
+            s += float(pmf_func(float(k)))
+        return float(np.clip(s, 0.0, 1.0))
```
Desiment: For distributions with infinite support, this method will not work. For example, if X, Y ~ Poiss(1), are independent, then X - Y has infinite support (that is there infinite amount of points in support.iter_leq(x)).

src/pysatl_core/distributions/fitters.py
```diff
@@ -338,3 +339,270 @@ def f(q: float) -> float:
...
+
+def fit_ppf_to_cdf_1D(
+    distribution: "Distribution", /, **options: Any
+) -> FittedComputationMethod[float, float]:
+    """
+    Fit **discrete** CDF using only a resolvable PPF via bisection on ``q``.
+
+    Semantics
+    ---------
+    ``CDF(x) = sup { q ∈ [0,1] : PPF(q) ≤ x }``
```
Desiment: Again, this depends on the definition of PPF, and how it is computed (as left or right). Note that this does not matter in contiuous case (or, at least, should not)

src/pysatl_core/distributions/distribution.py
LeonidElkin: Removed

PR #58, opened 2025-12-01
Families/normal family configuration

MyznikovFD:
- Implement Normal distribution with meanVar, meanPrec, exponential parameterizations
- Add analytical implementations for PDF, CDF, PPF, characteristic function
- Implement moment calculations (mean, variance, skewness, kurtosis)
- Configure family registry and parametrization system"

- Add tests covering all distribution characteristics
- Test multiple parameterizations and conversions
- Verify against scipy.stats.norm for correctness"

### DIFF
--- a/tests/unit/families/test_configuration.py
+++ b/tests/unit/families/test_configuration.py
@@ -1,0 +1,301 @@
+"""
+Tests for Normal Distribution Family Configuration
+
+This module tests the functionality of the normal distribution family
+defined in config.py, including parameterizations, characteristics,
+and sampling.
+"""
+
+__author__ = "Fedor Myznikov"
+__copyright__ = "Copyright (c) 2025 PySATL project"
+__license__ = "SPDX-License-Identifier: MIT"
+
+import math
+from typing import cast
+
+import numpy as np
+import pytest
+from scipy.stats import norm
+
+from pysatl_core.distributions.support import ContinuousSupport
+from pysatl_core.families.configuration import (
+    NormalExpParametrization,
+    NormalMeanPrecParametrization,
+    NormalMeanStdParametrization,
+    configure_families_register,
+)
+from pysatl_core.families.registry import ParametricFamilyRegister
+from pysatl_core.types import ContinuousSupportShape1D, UnivariateContinuous
+
+
+class TestNormalFamily:
+    """Test suite for Normal distribution family."""
+
+    # Precision for floating point comparisons
+    CALCULATION_PRECISION = 1e-10
+
+    def setup_method(self):
+        """Setup before each test method."""
+        registry = configure_families_register()
+        self.normal_family = registry.get("Normal Family")
+        self.normal_dist_example = self.normal_family(mu=2.0, sigma=1.5)
+
+    def test_family_registration(self):
+        """Test that normal family is properly registered."""
+        family = ParametricFamilyRegister.get("Normal Family")
+        assert family.name == "Normal Family"
+
+        # Check parameterizations
+        expected_parametrizations = {"meanStd", "meanPrec", "exponential"}
+        assert set(family.parametrization_names) == expected_parametrizations
+        assert family.base_parametrization_name == "meanStd"
+
+    def test_mean_var_parametrization_creation(self):
+        """Test creation of distribution with standard parametrization."""
+        dist = self.normal_family(mu=2.0, sigma=1.5)
+
+        assert dist.family_name == "Normal Family"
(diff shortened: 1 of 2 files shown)


src/pysatl_core/families/config.py
```diff
@@ -0,0 +1,343 @@
...
+
+def configure_family_register() -> None:
+    """
+    Configure and register all distribution families in the global registry.
+
+    This function initializes all parametric families with their respective
+    parameterizations, characteristics, and sampling strategies. It should be
+    called during application startup to make distributions available.
+    """
+    _configure_normal_family()
```
LeonidElkin: It is required `@lru_cache(maxsize=1)` decorator and returning `ParametricFamilyRegister` as well. Like in `distributions/registry.py`

```python
@lru_cache(maxsize=1)
def distribution_type_register() -> DistributionTypeRegister:
    reg = DistributionTypeRegister()
    _configure(reg)
    return reg
```

but you can do it like 

```python
@lru_cache(maxsize=1)
def configure_family_register() -> ParametricFamilyRegister:
    _configure_normal_family()
    return ParametricFamilyRegister()
```

I do like the style of using `ParametricFamilyRegister.register(Smth)`. It seems more natural in term of singletone object so you have no need in coding `reg = DistributionTypeRegister()` and pass it through the configuration function. I think I'll do the same thing in the new graph implementation. But we really need caching and returning the register object in such a function

src/pysatl_core/families/config.py
LeonidElkin: Rename this file to `configuration.py`

src/pysatl_core/families/registry.py
```diff
@@ -93,7 +93,26 @@ def register(cls, family: ParametricFamily) -> None:
...
     """Reset the cached distribution type register (test helper)."""
-    ParametricFamilyRegister._instance = None
+    ParametricFamilyRegister.clear()
+
+    try:
+        from pysatl_core.families.configuration import configure_family_register
+
+        configure_family_register.cache_clear()
+    except ImportError:
+        pass
```
LeonidElkin: Why are you using try-except here? Why no import it in the top of the file?

MyznikovFD: Importing in top of the file will entail cyclic import. Try-except prevents it

tests/unit/families/test_configuration.py
```diff
@@ -111,106 +108,106 @@ def test_pdf_calculation(self):
...
+
+    @pytest.mark.parametrize(
+        "char_func_getter, expected",
+        [
+            (lambda d: d.computation_strategy.query_method("mean", d)(None), 2.0),
+            (lambda d: d.computation_strategy.query_method("var", d)(None), 2.25),
+            (lambda d: d.computation_strategy.query_method("skewness", d)(None), 0.0),
+            (lambda d: d.computation_strategy.query_method("excess_kurtosis", d)(None), 0.0),
+        ],
+    )
```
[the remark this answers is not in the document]
LeonidElkin: Yeah, sure, delete it. We'll figure out what to do with this problem later

src/pysatl_core/families/configuration.py
```diff
@@ -0,0 +1,359 @@
...
+        Percent point function (inverse CDF) for normal distribution.
+
+        Parameters
+        ----------
+        parameters : Parametrization
+            Distribution parameters object with fields:
+            - mu: float (mean)
+            - sigma: float (standard deviation)
+        p : npt.NDArray[np.float64]
+            Probability from [0, 1]
```
Desiment: You should add to docstring behavior at p = 0 and p = 1

Desiment (changes requested):

src/pysatl_core/families/configuration.py
```diff
@@ -0,0 +1,359 @@
...
+
+
+PDF = "pdf"
+CDF = "cdf"
+PPF = "ppf"
+CF = "char_func"
+MEAN = "mean"
+VAR = "var"
+SKEW = "skewness"
+RAWKURT = "raw_kurtosis"
```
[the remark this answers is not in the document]
MyznikovFD: The solution will entail readability decrease, because the call: 
    `kurt_func = self.normal_dist_example.computation_strategy.query_method("kurtosis", ...)`
    `excess_kurt = kurt_func(True)`
is not clear and architect specificity is not allow `kurt_func(excess = True)`. So I prefer to leave it as it is

Desiment (approved):

PR #59, opened 2025-12-04
Feature/add kwargs to calculation functions

MyznikovFD:


tests/unit/distributions/test_basic.py
```diff
@@ -2,11 +2,17 @@
...
+
+from mypy_extensions import KwArg
 
 from pysatl_core.distributions.computation import AnalyticalComputation
 from pysatl_core.distributions.distribution import StandaloneEuclideanUnivariateDistribution
 from pysatl_core.types import Kind
 
+if TYPE_CHECKING:
+    pass
+
```
LeonidElkin: Don't import if you don't use it

LeonidElkin (approved):

PR #62, opened 2025-12-09
Refactor/naming convention

LeonidElkin:
Unified naming for:
- characteristics
- families
- parameterizations


MyznikovFD (changes requested):
I found now issues in the wrote code, but it does not correspond to the current main

PR #63, opened 2025-12-14
Families/exponential family configuration

MyznikovFD:
- Implemented an exponential parametric family

- File with implementations of concrete parametric families is separated into 3 files: normal.py, uniform.py, exponential.py

- New files are located in their own directory builtins/contitnuous

### DIFF
--- a/tests/unit/families/builtins/continuous/test_exponential.py
+++ b/tests/unit/families/builtins/continuous/test_exponential.py
@@ -1,0 +1,255 @@
+"""
+Tests for Exponential Distribution Family
+
+This module tests the functionality of the exponential distribution family,
+including parameterizations, characteristics, and sampling.
+"""
+
+__author__ = "Fedor Myznikov"
+__copyright__ = "Copyright (c) 2025 PySATL project"
+__license__ = "SPDX-License-Identifier: MIT"
+
+
+import numpy as np
+import pytest
+from scipy.stats import expon
+
+from pysatl_core.distributions.support import ContinuousSupport
+from pysatl_core.families.configuration import configure_families_register
+from pysatl_core.types import (
+    CharacteristicName,
+    ContinuousSupportShape1D,
+    FamilyName,
+    UnivariateContinuous,
+)
+
+from .base import BaseDistributionTest
+
+
+class TestExponentialFamily(BaseDistributionTest):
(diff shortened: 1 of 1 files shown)


src/pysatl_core/families/builtins/continuous/exponential.py
```diff
@@ -0,0 +1,270 @@
...
+            If probability is outside [0, 1]
+        """
+        if np.any((p < 0) | (p > 1)):
+            raise ValueError("Probability must be in [0, 1]")
+
+        parameters = cast(_Rate, parameters)
+        lambda_ = parameters.lambda_
+
+        with np.errstate(divide="ignore", invalid="ignore"):
+            return np.where(p < 1.0, -np.log(1.0 - p) / lambda_, np.inf)
```
LeonidElkin: -log(1-p)/lambda == -log1p(-p)/lambda, but log1p is more accurate near p=1

src/pysatl_core/families/builtins/continuous/exponential.py
```diff
@@ -0,0 +1,270 @@
...
+            Kurtosis value
+        """
+        if not excess:
+            return 9.0
+        else:
+            return 6.0
+
+    def _exponential_support(_: Parametrization) -> ContinuousSupport:
+        """Support of exponential distribution"""
+        return ContinuousSupport(left=0.0, right=math.inf)
```
LeonidElkin: `right=math.inf` is redundant here. We already have right=inf by default in `ContinuousSupport` .

tests/unit/families/builtins/continuous/test_exponential.py
```diff
@@ -0,0 +1,270 @@
...
+            CharacteristicName.PDF,
+            CharacteristicName.CDF,
+            CharacteristicName.PPF,
+            CharacteristicName.CF,
+            CharacteristicName.MEAN,
+            CharacteristicName.VAR,
+            CharacteristicName.SKEW,
+            CharacteristicName.KURT,
+        }
+        assert set(comp.keys()) == expected_chars
```
LeonidElkin: Here and in similar tests in other families, there is a relic of the past in the form of a naming. Caching is no longer checked here, we need to rename it to something that reflects the point.

Desiment (approved):

PR #67, opened 2026-02-03
Families/ adding base version of exponential family

### LINKED ISSUE #50: ExponentialClassFamily: conjugate prior support and posterior hyperparameters
# ExponentialClassFamily: conjugate prior support and posterior hyperparameters

## Task

Extend `ExponentialClassFamily` with a generic interface for conjugate priors and posterior hyperparameter updates.

### 1. Conjugate prior family property

Add a property:

- `conjugate_prior_family` (likely marked with `@property`),

which returns a `ParametricFamily` (or a subclass) representing the **conjugate prior** over the natural parameter (or over the model parameter, depending on design).

Expected shape (conceptually):

- if the model has density
  `f(x, θ) = h(x) * exp(<η(θ), T(x)> + A(θ))`,
- then the conjugate prior over the natural parameter `η` often has log-density of the form
  `log π(η | α, β) = <η, α> + β * A(η) + const`,
  where `α`, `β` are hyperparameters.

You do **not** need to fix a specific parametric form in this sub-issue, but:

- the `conjugate_prior_family` object must expose parameters (“hyperparameters”) that can be updated in a generic way in `posterior_hyperparameters`.

### 2. Hyperparameter representation

Introduce a small, explicit representation of hyperparameters for the conjugate prior, for example:

- a dataclass-like structure `ExponentialConjugateHyperparameters`,
- or a clearly documented dictionary schema (`{"alpha": ..., "beta": ...}`).

This structure will be the input and output of `posterior_hyperparameters`.

### 3. `posterior_hyperparameters` method

Add a method:

- `posterior_hyperparameters(prior_hyperparameters, sample)`,

which:

- takes:
  - `prior_hyperparameters` in the chosen representation,
  - `sample`, a collection of observations `x_1, ..., x_n`,
- returns:
  - updated hyperparameters corresponding to the posterior.

In general exponential-family theory, for canonical parametrization:

- the posterior hyperparameters are obtained by:
  - adding the **sum of sufficient statistics** to the “α-like” hyperparameter,
  - adding the **sample size** `n` to the “β-like” hyperparameter (or equivalent).

You should:

- implement this generic update rule in terms of:
  - `T(x)` (sufficient statistics),
  - the sample size `n = len(sample)`,
  - the prior hyperparameters.

Do **not** hard-code anything specific to a concrete family (Normal, Bernoulli, etc.) in this method.

domosedy:


src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,363 @@
...
+                theta,
+                self._log_partition(parametrization),
+            ]
+
+        def conjugate_log_partition(
+            parametrization: ExponentialFamilyParametrization,
+        ) -> Any:
+            alpha = parametrization.theta[0]
+            beta = parametrization.theta[1]
+
```
Desiment: Accept multivariate parameter

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,363 @@
...
+            parametrization = cast(ExponentialFamilyParametrization, parametrization)
+            return self._second_moment(parametrization, x) - self._mean(parametrization, x) ** 2
+
+        return func
+
+    def posterior_hyperparameters(
+        self, prior_hyper: ExponentialConjugateHyperparameters, sample: list[Any]
+    ) -> ExponentialConjugateHyperparameters:
+        alpha = prior_hyper.alpha
+        beta = prior_hyper.beta
```
Desiment: naming:

alpha -> effective_suff_stat_value
beta -> effective_sample_size

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,363 @@
...
+        SpacePredicate.__init__(self, lambda x: doesAccept(x, space))
+        self._space = space
+
+
+class NaturalExponentialFamily(ParametricFamily):
+    def __init__(
+        self,
+        *,
+        log_partition: Callable[[ExponentialFamilyParametrization], float],
+        sufficient_statistics: Callable[[Any], Any],
```
Desiment: Maybe enforce ``Callable[[npt.NDArray[np.float64]], npt.NDArray[np.float64]]``?

Desiment (changes requested):

src/pysatl_core/types.py
```diff
@@ -240,6 +240,25 @@ def shape(self) -> ContinuousSupportShape1D:
...
+            x = np.array([x])
+
+        return all(
+            x_coordinate in interval
+            for interval, x_coordinate in zip(self.intervals, x, strict=True)
+        )
+
+    def __contains__(self, x: object) -> bool:
+        """Check if a single point is in the interval."""
+        return bool(self.contains(cast(Number, x)))
```
LeonidElkin: Strange implementation of the `contains` method. If we check whether a point is contained in an N-dimensional interval, then at least we must accept an array of N elements, if we want to check M Points at once, then we must accept an array of M x N. You definitely don't need to implement the magic method `__contains__` here

src/pysatl_core/distributions/support.py
```diff
@@ -49,6 +56,15 @@ class ContinuousSupport(Interval1D, Support):
     """
 
 
+class ContinuousNDSupport(IntervalND, Support):  # type: ignore[misc]
```
LeonidElkin: Do we really need `# type: ignore[misc]` here?

src/pysatl_core/distributions/support.py
```diff
@@ -430,10 +446,26 @@ def is_right_bounded(self) -> bool:
...
+
+    def __contains__(self, item: NumericArray | Number) -> bool:
+        return self._predicate(item)
+
+
+class SupportByIntervals(SupportByPredicate):
+    def __init__(self, support: ContinuousNDSupport):
+        SupportByPredicate.__init__(self, lambda x: x in support)
+
+
```
LeonidElkin: Do we really need these wrappers? `SupportByPredicate` does not implement the `Support` protocol since it lacks `contains()`, and `SupportByIntervals` only delegates membership checks back to the wrapped interval support. It seems that these classes add an extra abstraction layer without providing new behavior, while making the support API less consistent.

src/pysatl_core/families/__init__.py
```diff
@@ -34,6 +40,10 @@
     "configure_families_register",
     # builtins
     *_builtins_all,
+    "ContinuousExponentialClassFamily",
+    "ExponentialFamilyParametrization",
+    "ExponentialConjugateHyperparameters",
+    # "CanonicalContinuousExponentialClassFamily",
```
LeonidElkin: What are those comments for?

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,309 @@
...
+    UnivariateContinuous,
+)
+
+if TYPE_CHECKING:
+    from pysatl_core.distributions.support import Support
+    from pysatl_core.types import Number, NumericArray
+
+    type ParametrizedFunction = Callable[[Parametrization, Any], Any]
+    type SupportArg = Callable[[Parametrization], Support | None] | None
+    type NumberParameter = Number | NumericArray
```
LeonidElkin: Move those types to `types.py`. `SupportArg` is already used in `parametric_family.py` so change there as well. `NumberParameter` is a bad naming, better smth like `ArrayLikeNumeric` or

src/pysatl_core/families/exponential_family.py
LeonidElkin, line 1: Docstring in the top of the file

LeonidElkin, line 1: Forgot to add such header as well

```python
__author__ = "Leonid Elkin"
__copyright__ = "Copyright (c) 2025 PySATL project"
__license__ = "SPDX-License-Identifier: MIT"
```

Of course, change with your name

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,309 @@
...
+    effective_suff_stat_value: NumericArray
+    effective_sample_size: Number
+
+    def transform_to_base_parametrization(self) -> ExponentialFamilyParametrization:
+        return ExponentialFamilyParametrization(
+            np.append(self.effective_suff_stat_value, self.effective_sample_size)
+        )
+
+
+class ContinuousExponentialClassFamily(ParametricFamily):
```
LeonidElkin, line 108: This class could be incompatible with `ParametricFamily` in main branch. Should rebase your branch

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,309 @@
...
+
+        ParametricFamily.__init__(
+            self,
+            name=name,
+            distr_type=distr_type,
+            distr_parametrizations=distr_parametrizations,
+            distr_characteristics=distr_characteristics,
+            support_by_parametrization=support_by_parametrization,
+        )
+        parametrization(family=self, name="theta")(ExponentialFamilyParametrization)
```
LeonidElkin: If we want the exponential class to be a certain narrowing of families with special properties, then we must accept user characteristics, and not nail down just these three

src/pysatl_core/families/exponential_family.py
LeonidElkin, line 1: Add docstring to every method in the file

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,309 @@
...
+            name=self.name,
+            distr_type=self._distr_type,
+            distr_parametrizations=self.parametrization_names,
+            support_by_parametrization=self.support_resolver,
+        )
+
+    def transform(
+        self,
+        transform_function: Callable[[Any], Any],
+    ) -> ContinuousExponentialClassFamily:
```
LeonidElkin, line 344: There is a strange method. `Callable[Any, Any]` literally tells nothing

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,309 @@
...
+        self._parameter_space = parameter_space
+        self._sufficient_statistics_values = sufficient_statistics_values
+
+        distr_characteristics: dict[
+            GenericCharacteristicName,
+            dict[ParametrizationName, ParametrizedFunction] | ParametrizedFunction,
+        ] = {
+            CharacteristicName.PDF: self.density,
+            CharacteristicName.MEAN: self._mean,
+            CharacteristicName.VAR: self._var,
```
LeonidElkin, line 185: I'm also not sure that these methods will always be more effective than the ones we add to builtins. In general, such an implementation of characteristics, in any case, should expand distr_characteristics, and not replace it. And expand it under other labels. Judging by the current pull request, you don't have any family labels in your branch yet, so you need to update.

tests/unit/families/test_exponential_family.py
```diff
@@ -0,0 +1,87 @@
...
+    conjugate_fam = fam.conjugate_prior_family.transform(transform_function)
+    ParametricFamilyRegister().register(conjugate_fam)
+    return cast(
+        ContinuousExponentialClassFamily,
+        ParametricFamilyRegister().get("TransformedExponentialFamily"),
+    )
+
+
+@pytest.mark.parametrize("theta1", range(2, 5))
+@pytest.mark.parametrize("theta2", range(2, 5))
```
LeonidElkin:
```suggestion
@pytest.mark.parametrize(
    ("theta1", "theta2"),
    itertools.product(range(2, 5), range(2, 5)),
)
```

and so on below

LeonidElkin (changes requested):

src/pysatl_core/types.py
domosedy: I think the case when we check M points at once isn't really needed.
`__contains__` method is mandatory for `in` operator

src/pysatl_core/families/exponential_family.py
domosedy: I added new fields in `CharacteristicName` class for class-defined implementation of these characteristics

src/pysatl_core/families/exponential_family.py
domosedy: This type is not ArrayLike it's one of Number or Array. So I really don't know how to name this type

domosedy: We mustn't accept Number

src/pysatl_core/types.py
domosedy: We should accept ND-array

src/pysatl_core/distributions/support.py
```diff
@@ -49,6 +62,16 @@ class ContinuousSupport(Interval1D, Support):
...
 
+# Support want to have Number as a parameter of contains, but we decided that we should avoid this
+class ContinuousNDSupport(IntervalND, Support):  # type: ignore[misc]
+    """
+    Support for continuous distributions represented as an array of intervals.
+
+    This class inherits from IntervalND and implements the Support protocol
+    for continuous distributions defined on a list of intervals [left, right].
+    """
+
```
LeonidElkin: The example of a multidimensional support showed the inconsistency of the original protocol. We need to change it. Remove overloading from the main `Support` protocol and make it work only with arrays. Well, accordingly, remove the `__contains__` methods everywhere in supports, since they do not work with arrays. More precisely, they work, but they won't be able to return the `BoolArray`

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,462 @@
...
+            if np.array([x]) not in self._support:
+                return -np.inf
+
+            theta = parametrization.theta
+            sufficient = self._sufficient(x)
+            dot = np.dot(theta, sufficient)
+            if hasattr(dot, "__len__"):
+                dot = dot[0]
+
+            result = np.log(self._normalization(x)) + dot + self._log_partition(theta)
```
LeonidElkin, line 245: The formula does not correspond to the docstring and standard notation I guess. 

wiki:

$$f(x \mid \theta) = h(x)\exp\left(\theta^\top T(x) - A(\theta)\right)$$

your code:

$$\log f(x \mid \theta)
= \log h(x) + \theta^\top T(x) + A(\theta)$$

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,462 @@
...
+        self._parameter_space = parameter_space
+        self._sufficient_statistics_values = sufficient_statistics_values
+
+        distr_characteristics: dict[
+            GenericCharacteristicName,
+            dict[ParametrizationName, ParametrizedFunction] | ParametrizedFunction,
+        ] = {
+            CharacteristicName.PDF: self.density,
+            CharacteristicName.MEAN_DEFAULT: self._mean,
+            CharacteristicName.VAR_DEFAULT: self._var,
```
LeonidElkin: Why do you put mean and var under the different names. It breaks the contract of our architecture. `ParametricFamily.__init__()` allow to put same characteristics under different labels. Come up with some kind of default label and put them under that label, but their name should be `CharacteristicName.MEAN` and `CharacteristicName.VAR`.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,462 @@
...
+
+    @property
+    def _var(self) -> ParametrizedFunction:
+        """Compute the variance Var[X] = E[X²] - (E[X])²."""
+
+        def func(parametrization: Parametrization, x: Any) -> Any:
+            parametrization = cast(ExponentialFamilyParametrization, parametrization)
+            return self._second_moment(parametrization, x) - self._mean(parametrization, x) ** 2
+
+        return func
```
LeonidElkin, line 444: These functions should not accept a dummy `x` argument, it was fixed a long time ago, and ur tests simply hide this problem by calling `mean(12)`

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,462 @@
...
+            sufficient_statistics=new_sufficient,
+            normalization_constant=new_normalization,
+            support=SupportByPredicate(predicate=new_support),
+            parameter_space=self._parameter_space,
+            sufficient_statistics_values=self._sufficient_statistics_values,
+            name=f"Transformed{self._name}",
+            distr_type=self._distr_type,
+            distr_parametrizations=self.parametrization_names,
+            support_by_parametrization=self.support_resolver,
+        )
```
LeonidElkin, line 396: **AI finding**

`transform(...)` is semantically incorrect or at least incorrectly documented: the docstring talks about the direct conversion of $y = g(x)$, and the code uses the function as an inverse. Also, the base measure should be considered at the starting point, and not in the transformed variable.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,462 @@
...
+
+        family = ParametricFamily(
+            name=f"PosteriorPredictive{self.name}",
+            distr_type=UnivariateContinuous,
+            distr_characteristics={CharacteristicName.PDF: posterior_density},
+            distr_parametrizations=["posterior"],
+            support_by_parametrization=lambda _: ContinuousSupport(),
+        )
+        parametrization(family=family, name="posterior")(ExponentialConjugateHyperparameters)
+        return family
```
LeonidElkin, line 540: Need more tests of it. Can't understand the correctness

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,462 @@
...
+            )
+            posterior_effective_sample_size += len(sample)
+        else:
+            posterior_effective_suff_stat_value += self._sufficient(sample)  # type: ignore[arg-type]
+            posterior_effective_sample_size += 1
+
+        return ExponentialConjugateHyperparameters(
+            effective_suff_stat_value=posterior_effective_suff_stat_value,
+            effective_sample_size=posterior_effective_sample_size,
+        )
```
LeonidElkin: Can we not use In-place array modification here?

src/pysatl_core/families/exponential_family.py
LeonidElkin: You don't allow the user to set their own characteristics, but you nail your own down. I'm not sure if this is justified in the context of the exponential class, but it seems wrong to me.

src/pysatl_core/distributions/support.py
```diff
@@ -430,10 +453,28 @@ def is_right_bounded(self) -> bool:
...
+    @overload
+    def contains(self, x: Number) -> bool: ...
+    @overload
+    def contains(self, x: NumericArray) -> BoolArray: ...
+
+    def contains(self, x: NumericArray) -> bool | BoolArray:  # type: ignore[misc]
+        return self.predicate(x)
+
+    def __contains__(self, item: object) -> bool | BoolArray:
+        return self.contains(cast(NumericArray, item))
```
LeonidElkin: According to the python data model, the magic method `__contains__` must always return `bool`. I would not define this method at all then.

tests/unit/families/test_exponential_family.py
LeonidElkin, line 1: There are not enough tests, only a narrow happy-path is checked. Moreover, in current tests, `RuntimeWarning: divide by zero encountered in log` occurs due to `np.log(-parametrization)` at the integration boundary. That's a bad bell.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,462 @@
...
+    def __init__(
+        self,
+        *,
+        log_partition: Callable[[NumericArray], NumericArray],
+        sufficient_statistics: Callable[[NumericArray], NumericArray],
+        normalization_constant: Callable[[NumericArray], Number],
+        support: SupportByPredicate,
+        parameter_space: SupportByPredicate,
+        sufficient_statistics_values: SupportByPredicate,
+        name: str = "ExponentialFamily",
```
LeonidElkin: `name` should not have default parameter

src/pysatl_core/distributions/support.py
LeonidElkin: New contract should look like this

```python
@runtime_checkable
class Support(Protocol):
    def contains(self, x: NumericArray) -> bool | BoolArray: ...
```

src/pysatl_core/distributions/support.py
LeonidElkin:
```suggestion
@dataclass(frozen=True, slots=True)
class PredicateSupport(Support):
    predicate: Callable[[NumericArray], bool | BoolArray]

    def contains(self, x: NumericArray) -> bool | BoolArray:
        return self.predicate(x)
```

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,462 @@
...
+
+    def __init__(
+        self,
+        *,
+        log_partition: Callable[[NumericArray], NumericArray],
+        sufficient_statistics: Callable[[NumericArray], NumericArray],
+        normalization_constant: Callable[[NumericArray], Number],
+        support: SupportByPredicate,
+        parameter_space: SupportByPredicate,
+        sufficient_statistics_values: SupportByPredicate,
```
LeonidElkin: Too strict signature. Change to 

```python
support: Support,
parameter_space: Support,
sufficient_statistics_values: Support,
```

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,462 @@
...
+        ParametricFamily.__init__(
+            self,
+            name=name,
+            distr_type=distr_type,
+            distr_parametrizations=distr_parametrizations,
+            distr_characteristics=distr_characteristics,
+            support_by_parametrization=support_by_parametrization,
+            base_score=base_score,
+        )
+        parametrization(family=self, name="theta")(ExponentialFamilyParametrization)
```
LeonidElkin: **AI finding**

`parameter_space` is stored but never used to validate `theta`.

`ExponentialFamilyParametrization` has no constraint, so `ParametricFamily.distribution()` accepts natural parameters outside the declared parameter space. Those invalid parameters then reach `log_partition(theta)` in `log_density`, where they can produce `nan`/`inf` instead of failing at construction time.

The `theta` parametrization should validate that `theta` belongs to `parameter_space` before a distribution instance is created.

src/pysatl_core/families/exponential_family.py
domosedy: Yes, but I decided that + log-partition function is easier to write and check formulas. This is described in docstring of class

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,483 @@
...
+
+        family = ParametricFamily(
+            name=f"PosteriorPredictive{self.name}",
+            distr_type=UnivariateContinuous,
+            distr_characteristics={CharacteristicName.PDF: posterior_density},
+            distr_parametrizations=["posterior"],
+            support_by_parametrization=lambda _: ContinuousSupport(),
+        )
+        parametrization(family=family, name="posterior")(ExponentialConjugateHyperparameters)
+        return family
```
LeonidElkin, line 540: **AI finding**

`posterior_predictive` does not preserve the original observation support.

The predictive density is defined for a new observation `x`, so it should be zero outside the original family support. Currently `posterior_density` does not check `self._support`, and the created family uses `ContinuousSupport()` as an unconditional real-line support.

For example, for the exponential-family test case with support `x > 0`, `pdf(-0.5)` currently returns a positive value, but it should be outside support.

The posterior predictive family should reuse the original support (or a proper support resolver), and `posterior_density` should return `0` outside it.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,483 @@
...
+            sufficient_statistics=new_sufficient,
+            normalization_constant=new_normalization,
+            support=PredicateSupport(predicate=new_support),
+            parameter_space=self._parameter_space,
+            sufficient_statistics_values=self._sufficient_statistics_values,
+            name=f"Transformed{self._name}",
+            distr_type=self._distr_type,
+            distr_parametrizations=self.parametrization_names,
+            support_by_parametrization=self.support_resolver,
+        )
```
LeonidElkin, line 396: I might be wrong but the transformed base measure is still evaluated at the wrong point.

The docstring now treats `transform_function` as the inverse map:

$$
x = g(y)
$$

For a change of variables, the transformed density should use:

$$
p_Y(y) = h(g(y)) \exp(\theta^\top T(g(y)) + A(\theta)) \left|J_g(y)\right|
$$

but `new_normalization` currently computes:

$$
h(y) \left|J_g(y)\right|
$$

because it calls `self._normalization(x)` instead of `self._normalization(transform_function(x))`.

This is hidden by current tests because they use `h(x) = 1`. Add a test with a non-constant base measure.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,483 @@
...
+        self._parameter_space = parameter_space
+        self._sufficient_statistics_values = sufficient_statistics_values
+
+        family_characteristics: CharacteristicsMap = {
+            CharacteristicName.PDF: self.density,
+            CharacteristicName.MEAN: self._mean,
+            CharacteristicName.VAR: self._var,
+        }
+        merged_characteristics = dict(distr_characteristics or {})
+        merged_characteristics.update(family_characteristics)
```
LeonidElkin: The merge order gives default characteristics higher priority than user-provided ones.

`distr_characteristics` is copied first, then `family_characteristics` is applied with `update()`, so user-provided `PDF`, `MEAN`, or `VAR` are silently overwritten by the generic implementations.

This makes it impossible to replace expensive numerical `MEAN` / `VAR` with analytical implementations for a concrete exponential family. Use labels so the generic implementation does not overwrite the analytical one.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,483 @@
...
+            support_by_parametrization=support_by_parametrization,
+            base_score=base_score,
+        )
+
+        @parametrization(family=self, name="theta")
+        class ThetaParametrization(ExponentialFamilyParametrization):
+            @constraint(description="theta belongs to parameter_space")
+            def check_theta_in_parameter_space(self) -> bool:
+                theta = np.atleast_1d(np.asarray(self.theta, dtype=float))
+                return bool(self.__family__._parameter_space.contains(theta))  # type: ignore[attr-defined]
```
LeonidElkin: The `theta` validation is coupled to a private family attribute and assumes scalar membership.

`check_theta_in_parameter_space` reaches into `self.__family__._parameter_space` and then casts the result to `bool`. This is fragile for two reasons:

1. it depends on a private attribute and requires `type: ignore`;
2. `Support.contains()` may return a `BoolArray`, and `bool(array)` fails for arrays with more than one element.

Route this through a small helper that explicitly handles scalar-vs-batch membership, or make the parameter-space check return a guaranteed scalar boolean for a single `theta`.

tests/unit/distributions/test_support.py
LeonidElkin, line 1: The new `Support` protocol is array-only, but tests still exercise scalar calls directly.

The protocol now says:

```python
def contains(self, x: NumericArray) -> bool | BoolArray: ...
```

but several tests still call `contains(point)` with plain Python scalars. Runtime works because implementations call `np.asarray`, but the tests no longer reflect the declared API contract.

Update tests and internal call sites to pass `np.asarray(...)` / arrays consistently, or change the protocol if scalar inputs are still intended to be supported.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,483 @@
...
+            support: Predicate defining the support of the distribution.
+            parameter_space: Predicate defining the natural parameter space.
+            sufficient_statistics_values: Predicate defining the range of T(x).
+            name: Name of the family.
+            distr_type: Type of distribution or a callable returning it.
+            distr_parametrizations: List of parametrization names this family supports.
+            distr_characteristics: Additional analytical characteristics to register.
+            support_by_parametrization: Callable that returns the support given a parametrization.
+            base_score: Optional base score function.
+        """
```
LeonidElkin, line 173: Docstrings in this module should be aligned with the project style and the actual formulas. We require NumPy-style docstrings, but this constructor uses `Args:`. There are also inconsistencies around the meaning of `log_partition`, and a typo in `parametrizaiton`.

Please rewrite the public method docstrings in NumPy style and make the mathematical notation consistent with the implementation.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,483 @@
...
+            n = n₀ + N
+
+        Args:
+            parametrizaiton: Current conjugate hyperparameters.
+            sample: List of observations (each can be scalar or array).
+
+        Returns:
+            ExponentialConjugateHyperparameters:
+                Updated hyperparameters after incorporating the sample.
+        """
```
LeonidElkin: Same docstring issues here

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,540 @@
...
+            sufficient_statistics=conjugate_sufficient,
+            normalization_constant=lambda _: 1,
+            support=self._parameter_space,
+            sufficient_statistics_values=self._parameter_space,
+            parameter_space=PredicateSupport(predicate=conjugate_sufficient_accepts),
+            name=self.name,
+            distr_type=self._distr_type,
+            distr_parametrizations=self.parametrization_names,
+            support_by_parametrization=self.support_resolver,
+        )
```
LeonidElkin, line 339: `_support` is changed for the conjugate/transformed family, but the family still reuses `self.support_resolver`.

This makes the density use the new support while `distribution.support` can remain `None` or point to the original support. Expose the same transformed support through `support_by_parametrization` as well, e.g. for `transform()` return a resolver for `PredicateSupport(new_support)` and for the conjugate prior return the natural-parameter support.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,540 @@
...
+            sufficient_statistics=new_sufficient,
+            normalization_constant=new_normalization,
+            support=PredicateSupport(predicate=new_support),
+            parameter_space=self._parameter_space,
+            sufficient_statistics_values=self._sufficient_statistics_values,
+            name=f"Transformed{self._name}",
+            distr_type=self._distr_type,
+            distr_parametrizations=self.parametrization_names,
+            support_by_parametrization=self.support_resolver,
+        )
```
LeonidElkin, line 396: Same issue as in comment above

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,540 @@
...
+        else:
+            posterior_effective_suff_stat_value = np.array(
+                parametrization.effective_suff_stat_value,
+            ) + np.asarray(self._sufficient(sample))  # type: ignore[arg-type]
+            posterior_effective_sample_size = parametrization.effective_sample_size + 1
+
+        return ExponentialConjugateHyperparameters(
+            effective_suff_stat_value=posterior_effective_suff_stat_value,
+            effective_sample_size=posterior_effective_sample_size,
+        )
```
LeonidElkin, line 489: **AI finding**

`hasattr(sample, "__iter__")` misclassifies zero-dimensional NumPy arrays as samples.

`posterior_hyperparameters(..., sample=np.asarray(0.5))` enters the iterable branch and raises `TypeError: iteration over a 0-d array`. Normalize the input explicitly, e.g. distinguish scalar/0-D observations from batches with `np.ndim(sample)`.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,540 @@
...
+        def posterior_density(parametrization: Parametrization, x: NumericArray) -> Number:
+            parametrization = cast(ExponentialConjugateHyperparameters, parametrization)
+            if not self._support.contains(np.asarray(x)):
+                return cast(np.float32, 0.0)
+            return cast(
+                np.float32,
+                self._normalization(x)
+                * conjugate_log_partition(parametrization)
+                / conjugate_log_partition(
+                    self.posterior_hyperparameters(parametrization=parametrization, sample=[x])
```
LeonidElkin, line 528: **AI finding**

`posterior_density` recomputes the conjugate-prior family and its normalizing integral on every density evaluation.

`conjugate_log_partition(...)` accesses `self.conjugate_prior_family`, and each call eventually runs `nquad`; `posterior_density` calls it twice per `x`. Construct/cache the conjugate family once in `posterior_predictive` and avoid recomputing invariant normalizers for every point.

src/pysatl_core/families/exponential_family.py
```diff
@@ -0,0 +1,540 @@
...
+        self._sufficient_statistics_values = sufficient_statistics_values
+
+        family_characteristics: CharacteristicsMap = {
+            CharacteristicName.PDF: self.density,
+            CharacteristicName.MEAN: self._mean,
+            CharacteristicName.VAR: self._var,
+        }
+        distr_characteristics = dict(distr_characteristics or {})
+        merged_characteristics = dict(family_characteristics)
+        merged_characteristics.update(distr_characteristics)
```
LeonidElkin, line 189: The constructor now preserves user characteristics, but the merge is still shallow and does not keep the generic exponential-family methods under separate labels.

If a user provides `MEAN`/`VAR`/`PDF`, the generic fallback is replaced entirely. The intended contract is to "expand under other labels". Merge by label and add a test for the collision case, not only for adding `CDF`.

src/pysatl_core/families/exponential_family.py
LeonidElkin: Sure

LeonidElkin (changes requested):

PR #76, opened 2026-03-03
Fitters.py refactoring

### LINKED ISSUE #26: Fitters.py refactoring
# Problem

The fitters are in extremely poor condition right now:
- They don't have enough different options to control
- They are not sufficiently tested
- Their set is extremely scarce
- They don't use the caching feature
- Еhey are coded incomprehensibly and in a hurry
- They are poorly optimized

# What needs to be done 

You need to review all the functions in the file and either replace them completely or bring them to an acceptable state

Everything should be perfectly tested

siwreienta:


src/pysatl_core/distributions/computations/__init__.py
LeonidElkin, line 1: Firstly, add a docstring with details of this submodule in the top of the file. Secondly, in order not to stretch the init files, I recommend making a scheme with importing __all__ from the files of this module. For details, see how it's done [here](https://github.com/PySATL/pysatl-core/blob/main/src/pysatl_core/\_\_init\_\_.py)

LeonidElkin: Actually, you need to add docstrings everywhere. I won't mention it farther

src/pysatl_core/distributions/computations/__init__.py
```diff
@@ -0,0 +1,70 @@
+from __future__ import annotations
+
+__author__ = "Leonid Elkin, Mikhail Mikhailov, Irina Sergeeva"
```
LeonidElkin: I think you can add only yourself to `author`

src/pysatl_core/distributions/fitters/base.py
```diff
@@ -0,0 +1,190 @@
...
+        ``False`` — the callable is a **direct evaluator** (``eval_`` prefix).
+        It is lightweight and called on every query without caching.
+
+        This flag determines which slot (``fitter`` vs ``evaluator``) is
+        populated when :meth:`to_computation_method` builds a
+        :class:`ComputationMethod`.
+    options : tuple[FitterOption, ...]
+        Structured option schema.
+    tags : frozenset[str]
+        Constraint tags used for matching (e.g. ``{"continuous", "univariate"}``).
```
LeonidElkin: Naming is not very telling. At least `constraint_tags` or smth like that

src/pysatl_core/distributions/computations/base.py
```diff
@@ -0,0 +1,190 @@
...
+    priority : int
+        Higher priority wins when multiple fitters match.  Default is 0.
+    description : str
+        Human-readable summary of what the fitter does.
+    """
+
+    name: str
+    target: GenericCharacteristicName
+    sources: Sequence[GenericCharacteristicName]
+    fitter: FitterFunc
```
LeonidElkin: We've already have such type in `computations.py`. Move it to `types.py` and use in both places

src/pysatl_core/distributions/fitters/base.py
```diff
@@ -0,0 +1,190 @@
...
+            return ComputationMethod(
+                target=self.target,
+                sources=list(self.sources),
+                fitter=self.fitter,
+            )
+        return ComputationMethod(
+            target=self.target,
+            sources=list(self.sources),
+            evaluator=self.fitter,  # type: ignore[arg-type]
+        )
```
LeonidElkin: It is a bit incorrect. You provide `FitterFunc` type to evaluator. I think it's worth doing the following: if the fitter is not cacheable, then we don't name it a `fitter`, but an `evaluator`. And we return from it not the `FittedComputationMethod`, but the already calculated value. This will help us maintain the overall consistency of the API. 

I think it's possible to rename submodule `fitters` to `computations` and handle both evaluator and fitters cases. You probably could move `computation.py` there

src/pysatl_core/distributions/computations/base.py
```diff
@@ -0,0 +1,190 @@
...
+        raw = kwargs.pop(self.name, self.default)
+        try:
+            value = self.type(raw)
+        except (TypeError, ValueError) as exc:
+            raise TypeError(
+                f"Option '{self.name}': cannot convert {raw!r} to {self.type.__name__}"
+            ) from exc
+        if self.validate is not None and not self.validate(value):
+            raise ValueError(f"Option '{self.name}': value {value!r} failed validation.")
+        return value
```
LeonidElkin: Based on our dialogue in an online call. The kwargs will have to leave, and we will use the descriptors to form kwargs for specific fitters and evaluators.

src/pysatl_core/distributions/computations/base.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+
+    name: str
+    target: GenericCharacteristicName
+    sources: Sequence[GenericCharacteristicName]
+    fitter: FitterFunc
+    cacheable: bool = True
+    options: tuple[FitterOption, ...] = ()
+    tags: frozenset[str] = field(default_factory=frozenset)
+    priority: int = 0
```
LeonidElkin: It doesn't seem that priority is used for smth. I think we should either remove it or change the strategy to consider it while  finding the path in the graph

src/pysatl_core/distributions/fitters/base.py
LeonidElkin: After it's done I think the cacheable field will be gone

src/pysatl_core/distributions/fitters/continuous.py
```diff
@@ -0,0 +1,376 @@
...
+        Must expose a ``pdf`` characteristic.
+    **kwargs : Any
+        ``limit`` (int, default 200) — max ``quad`` subdivisions.
+
+    Returns
+    -------
+    FittedComputationMethod[NumericArray, NumericArray]
+        Array-semantic ``cdf`` callable.
+    """
+    opts = FITTER_PDF_TO_CDF_1C.resolve_options(kwargs)
```
LeonidElkin: This is also should be gone after moving descriptors to strategy

src/pysatl_core/distributions/computations/continuous.py
```diff
@@ -0,0 +1,376 @@
...
+            type=int,
+            default=200,
+            description="Maximum number of quad subdivisions per integral.",
+            validate=lambda v: v > 0,
+        ),
+    ),
+    tags=frozenset({"continuous", "univariate"}),
+    priority=0,
+    description="PDF -> CDF via segment-wise scipy.integrate.quad with cumsum.",
+)
```
LeonidElkin: I think we should fill in the descriptors lazily when the strategy does not find an analytically defined characteristic in the graph and goes looking for paths in the graph, or when the user wants to get/change options in the descriptors. 

This is necessary so that when working with our builtins distributions in a standard way, you do not need to keep extra information in memory.

LeonidElkin: So I think it's worth keeping the descriptor padding and calculations in different places.

LeonidElkin (changes requested):
* Universal options (calculations) of the characteristic should be passed to `query_method` and stick out explicitly for the user as a public API instead of the current `**options: Any`. @Desiment  will describe the use cases in more detail in the PR or personally
* The strategy assumes responsibility for passing options to `ComputationMethod`. If no changes have been made by the user, we leave the defaults.

src/pysatl_core/distributions/computations/base.py
```diff
@@ -0,0 +1,452 @@
...
+    fitter: FitterFunc
+    characteristic_options: tuple[CharacteristicOption, ...] = ()
+    computation_options: tuple[ComputationOption, ...] = ()
+    constraint_tags: frozenset[str] = field(default_factory=frozenset)
+    description: str = ""
+
+    @property
+    def options(self) -> tuple[_BaseOption, ...]:
+        """All options (characteristic first, then computation)."""
+        return (*self.characteristic_options, *self.computation_options)
```
LeonidElkin: Remove redundant parentheses

src/pysatl_core/distributions/computations/base.py
```diff
@@ -0,0 +1,452 @@
...
+    """Resolve a tuple of option descriptors from *kwargs* (mutates *kwargs*)."""
+    return {opt.name: opt.resolve(kwargs) for opt in options}
+
+
+def _option_names(options: tuple[_BaseOption, ...]) -> tuple[str, ...]:
+    return tuple(opt.name for opt in options)
+
+
+def _option_defaults(options: tuple[_BaseOption, ...]) -> dict[str, Any]:
+    return {opt.name: opt.default for opt in options}
```
LeonidElkin: It's better to create some abstract Descriptor and move those functions there as methods

LeonidElkin: So on we can move some common fields there. A lot of functionality in Fitter and Evaluate descriptors are doubled

src/pysatl_core/distributions/computations/_utils.py
LeonidElkin, line 1: It's better to name this file `_utils.py` with a prefixed underscore so we don't show these functions in the public API

src/pysatl_core/distributions/computation.py
LeonidElkin, line 1: There is NO such thing as "backward-compatibility"  before release. Delete this file

src/pysatl_core/distributions/computations/registry.py
LeonidElkin, line 1: Let's talk a little bit about this class. Now it is not entirely clear why it is needed at all. 

There is the following contradiction here. If the descriptors are located in a public API and are initialized at the import level, then why do we need a registry at all? And so on `configuration.py` is not lazily configurated at all

I suggest two options: either we remove this class altogether, or we hide all the fitters and descriptors from the public API, make the registry a singleton, and create a configuration function that fills it with descriptors. In this case, the configuration function returns the registry, and simply creating it via init produces an empty registry. In this case, the configuration should be done logically in the graph configuration.

The second option is preferable, because that's how transformations and families work now.

src/pysatl_core/distributions/strategies.py
LeonidElkin, line 1: I think caching in the strategy would not work at all now. Probably it would return the method with the same options, not the new one. We should consider options when caching

tests/unit/distributions/computations/test_base.py
LeonidElkin, line 1: Divide it into several files instead of dividing different logic with comments

LeonidElkin (changes requested):

src/pysatl_core/distributions/computations/_utils.py
```diff
@@ -0,0 +1,290 @@
...
+    RuntimeError
+        If the distribution cannot provide the requested characteristic.
+    """
+    try:
+        fn = distribution.query_method(name)
+    except AttributeError as exc:
+        raise RuntimeError(
+            f"Distribution does not expose characteristic '{name}' "
+            "via computation_strategy.query_method()."
+        ) from exc
```
LeonidElkin: I don't think it's a good idea to mask the original exception with another one of your own

src/pysatl_core/distributions/computations/_utils.py
```diff
@@ -0,0 +1,290 @@
...
+            raise TypeError
+    except (TypeError, ValueError):
+        _fn_scalar = fn
+
+        def _vectorised(x: NumericArray, **kwargs: Any) -> NumericArray:
+            return cast("NumericArray", np.vectorize(lambda xi: _fn_scalar(float(xi), **kwargs))(x))
+
+        return _vectorised
+
+    return fn
```
LeonidElkin: In general, this approach, when we adjust the analytical characteristics to the array semantics, is kinda ok. However, I think it's worth explicitly stating this, for example, in the distribution docstring, that the recommended approach is to create an analytically defined characteristic through array semantics. Because it might not be completely trivial.

src/pysatl_core/distributions/computations/_utils.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def collect_support(support: DiscreteSupport) -> np.ndarray:
```
LeonidElkin: Rename to `collect_discrete_support`

src/pysatl_core/distributions/computations/_utils.py
```diff
@@ -0,0 +1,290 @@
...
+            cur = support.first()
+            seen: set[Any] = set()
+            while cur is not None and cur not in seen:
+                seen.add(cur)
+                xs.append(float(cur))
+                cur = support.next(cur)
+            if xs:
+                return np.array(sorted(xs), dtype=float)
+        except Exception:
+            pass
```
LeonidElkin: This fallback is kinda dead code for our implementation. Delete it so. It does not imply a custom extension by inheriting from the class `DiscreteSupport`

tests/unit/distributions/computations/test_helpers.py
LeonidElkin, line 1: No tests for `build_tail_table()` func

src/pysatl_core/distributions/computations/_utils.py
```diff
@@ -0,0 +1,290 @@
...
+    tail_from : np.ndarray
+        Tail probabilities, shape ``(m + 1,)``.
+        ``tail_from[i] = P(X >= xs[i])``, normalised so ``tail_from[0] == 1``.
+    """
+    max_k = support.max_k
+    assert max_k is not None, "build_tail_table requires support.max_k to be set"
+    residue = support.residue
+    modulus = support.modulus
+
+    xs = np.arange(residue, max_k + 1, modulus, dtype=float)
```
LeonidElkin: It seems incorrect. I'm not really sure about it but all $$x < residue$$ will cause $$CDF(x) = 0$$. @Desiment , ur thoughts?

src/pysatl_core/distributions/computations/descriptors.py
LeonidElkin, line 1: Let's divide this file to `options.py` and `descriptors.py`

src/pysatl_core/distributions/computations/base.py
```diff
@@ -0,0 +1,459 @@
...
+
+    Examples: ``max_iter`` (bisection iterations), ``x_tol`` (bracket
+    tolerance), ``limit`` (quad subdivisions), ``n_q_grid`` (PPF grid size).
+
+    These options are declared in ``FitterDescriptor.computation_options``.
+    """
+
+
+@dataclass(frozen=True, slots=True)
+class OptionsDescriptor:
```
LeonidElkin: Rename to ` EdgeOptionsDescriptor`

src/pysatl_core/distributions/registry/configuration.py
```diff
@@ -115,17 +142,31 @@ def _configure(reg: CharacteristicRegistry) -> None:
...
+        fitter_reg,
+        pairs=(
+            (CharacteristicName.PMF, CharacteristicName.CDF),
+            (CharacteristicName.CDF, CharacteristicName.PMF),
+            (CharacteristicName.CDF, CharacteristicName.PPF),
+            (CharacteristicName.PPF, CharacteristicName.CDF),
+        ),
+        tags=_DISCRETE_1D_TAGS,
+        constraint=edge_disc_dim1,
+    )
```
LeonidElkin, line 163: Okay, what's really going on here. We create a registry object, fill it with descriptors, which were already created during imports, then we made an auxiliary function to do `find()` on this registry itself. It just doesn't make sense.

We discussed at the conference that it is necessary to make the `FittersRegistry` a singleton, not to create descriptors when importing, but to create them lazily when configuring the graph. Also, all fitters must be named with a prefixed underscore so that they are not a public API.

src/pysatl_core/distributions/registry/configuration.py
```diff
@@ -4,6 +4,15 @@
  - No auto-configuration in constructor.
  - Provide ``characteristic_registry()`` with ``@lru_cache`` that builds the
   singleton instance and seeds it with a set of edges.
+
+Descriptors are registered lazily: only their metadata (target, sources, tags)
+is used to declare graph edges at configuration time.  The actual
+``to_computation_method()`` call happens on demand when the strategy resolves a path.
```
LeonidElkin: It is not true

src/pysatl_core/distributions/computations/discrete.py
```diff
@@ -0,0 +1,389 @@
...
+__all__ = [
+    "fit_pmf_to_cdf_1D",
+    "FITTER_PMF_TO_CDF_1D",
+    "fit_cdf_to_pmf_1D",
+    "FITTER_CDF_TO_PMF_1D",
+    "fit_cdf_to_ppf_1D",
+    "FITTER_CDF_TO_PPF_1D",
+    "fit_ppf_to_cdf_1D",
+    "FITTER_PPF_TO_CDF_1D",
+]
```
LeonidElkin: Do not include fitter functions to public API

src/pysatl_core/distributions/strategies.py
```diff
@@ -7,27 +7,172 @@
...
     from pysatl_core.types import GenericCharacteristicName, LabelName
 
 
+# --------------------------------------------------------------------------- #
+# Execution plan (introspection of how a strategy will compute a state)
+# --------------------------------------------------------------------------- #
+
+
+@dataclass(frozen=True, slots=True)
+class ExecutionStep:
```
LeonidElkin: Rename `ExecutionStep` to `ComputationStep` and `ExecutionPlan` to `ComputationPlan`

src/pysatl_core/distributions/distribution.py
```diff
@@ -291,6 +292,30 @@ def query_method(
         """
         return self.computation_strategy.query_method(characteristic_name, self, **options)
 
+    def explain(self, characteristic_name: GenericCharacteristicName) -> ExecutionPlan:
```
LeonidElkin: Rename to `explain_computation_path`

src/pysatl_core/distributions/strategies.py
```diff
@@ -168,65 +474,46 @@ def query_method(
...
+                else:
+                    fitted = method.prepare(distr, **resolved)
+                    if self._enable_caching and method.cacheable:
+                        self._cache[cache_key] = fitted
+
+                last_fitted = fitted
+
+            if last_fitted is None:
+                raise RuntimeError(f"Empty path when resolving '{state}'.")
+            return last_fitted
```
LeonidElkin, line 788: As far as I can see, the problem of duplicate names in options has not been solved yet. That is, if we pass a kwarg with the same name and there are two different options under this name, then they will both take this value.

In general, again, as mentioned earlier. We need to get rid of `**options` in `query_method`. It's just these `OptionDescriptor`'s that need to be passed there. 

By the way, you can add a method to these graph descriptors that will allow you to change the options and return a copy of the descriptor object. This way, the user can get a `ExecutionPlan`, see which options they need, change these options using the new method, and run `query_method` . Then it will be easier to make benchmarks of all kinds.

src/pysatl_core/distributions/computations/continuous.py
```diff
@@ -0,0 +1,445 @@
...
+            result[interior] = np.clip(np.frompyfunc(_single, 1, 1)(x_in).astype(float), 0.0, 1.0)
+
+        return maybe_unwrap_scalar(result)
+
+    return FittedComputationMethod[NumericArray, NumericArray](
+        target=CharacteristicName.CDF,
+        sources=[CharacteristicName.PPF],
+        func=_cdf,  # type: ignore[arg-type]
+    )
+
```
LeonidElkin: This fitter is returning `nan` on the bounds and out of bounds of the support. The previous version "clamped" the left in 0.0 and the right in 1.0. @Desiment am i right?

src/pysatl_core/distributions/computations/_utils.py
```diff
@@ -0,0 +1,290 @@
...
+        1-D result array.
+
+    Returns
+    -------
+    NumericArray
+        Scalar or array.
+    """
+    if result.shape == (1,):
+        return cast("NumericArray", result[0])
+    return cast("NumericArray", result)
```
LeonidElkin: Why do we turn an array into a scalar for a single-element array? We always accept and return arrays.

src/pysatl_core/distributions/strategies.py
```diff
@@ -168,65 +474,46 @@ def query_method(
...
+
+                cache_key = (id(distr), id(edge), method.target, _freeze_options(resolved))
+                cached_fitted: FittedComputationMethod[Any, Any] | None = None
+                if self._enable_caching:
+                    cached_fitted = self._cache.get(cache_key)
+
+                if cached_fitted is not None:
+                    fitted = cached_fitted
+                else:
+                    fitted = method.prepare(distr, **resolved)
```
LeonidElkin: **NOTICE: THIS ISSUE WAS FOUND WITH THE HELP OF AI AND A COMMENT AS WELL**

Issue: multi-hop conversion paths do not compose intermediate fitted methods

For a path with multiple edges, e.g. `pdf -> cdf -> ppf`, the strategy prepares every edge against the original `distribution` and only keeps the last fitted method in `last_fitted`.

The fitted result of the previous edge is never exposed as the source method for the next edge. This means the first edge is effectively prepared and then discarded unless it happens to be requested later through the distribution cache.

This becomes observable when earlier edges have options. For example, if `pdf -> cdf` is fitted with `cdf_bias=0.25`, the next `cdf -> ppf` fitter will typically call `distribution.query_method("cdf")` internally. That call resolves `cdf` again from the original distribution, without the original `cdf_bias`, so the final `ppf` is built from a different CDF than the one selected by the execution plan.

The current tests mostly verify fit-call counts and option-aware cache keys, but they do not check that the second edge actually consumes the fitted output of the first edge.

Suggested coverage: add a real two-hop test where the second fitter calls `distribution.query_method("cdf")`, and assert that the final result depends on an option declared on the first edge.

src/pysatl_core/distributions/computations/computation.py
```diff
@@ -77,153 +61,210 @@ def __call__(self, **options: Any) -> Out: ...
...
+        This allows evaluator-based methods to be used in the same way
+        as fitter-based methods when needed.
         """
-        if self.fitter is not None:
-            return self.fitter(distribution, **options)
 
-        def _bound(*args: Any, **kwargs: Any) -> Out:
-            return cast(Evaluator[In, Out], self.evaluator)(distribution, *args, **kwargs)
+        def _bound(*args: Any, **kwargs: Any) -> NumericArray:
+            return self.evaluator(distribution, *args, **kwargs)
```
LeonidElkin: **NOTICE: THIS ISSUE WAS FOUND WITH THE HELP OF AI AND A COMMENT AS WELL**

Issue: options passed to `EvaluatorMethod.prepare()` are dropped

`EvaluatorMethod.prepare()` accepts `**options`, and the strategy does pass resolved edge options into `method.prepare(distr, **resolved)`. However, the bound wrapper created here ignores those prepared options and forwards only the call-time `kwargs`.

As a result, this does not work as expected:

```python
cdf = distribution.query_method("cdf", tol=1e-6)
cdf(0.5)

**Review by LeonidElkin — COMMENTED** [review]
It's much better than it was before, but there are still a couple of issues that were discussed but not resolved.

Also, don't forget to make jupiter notebook in the `examples/` folder

**Review by Desiment — DISMISSED** [review]

File: src/pysatl_core/distributions/registry/configuration.py (hunk 17 above)
Comment (LeonidElkin) [inline src/pysatl_core/distributions/registry/configuration.py:163]: If you make correct singletone semantic, there will be no need to pass fitter_reg to the _add_edges function

File: src/pysatl_core/distributions/computations/registry.py [hunk 27]
```diff
@@ -0,0 +1,181 @@
...
+    >>> registry.register(some_descriptor)
+    >>> desc = registry.find("cdf", ["pdf"], required_tags={"continuous", "univariate"})
+    """
+
+    def __init__(self) -> None:
+        self._by_key: dict[
+            tuple[GenericCharacteristicName, tuple[GenericCharacteristicName, ...]],
+            list[FitterDescriptor],
+        ] = {}
+        self._all: list[FitterDescriptor] = []
```
LeonidElkin, line 57: Make correct singletone semantic. For example

```python
reg1 = fitter_registry()
reg2 = FitterRegistry()

reg1 is not reg2  # not the same object
```
Look carefully how it's implemented in families or distributions.registry modules

src/pysatl_core/distributions/computations/continuous.py
```diff
@@ -0,0 +1,452 @@
...
+
+def _fit_ppf_to_cdf_1C(
+    distribution: Distribution,
+    /,
+    q_lowest: float = 1e-12,
+    q_highest: float = 1.0 - 1e-12,
+    max_iter: int = 256,
+) -> FittedComputationMethod[NumericArray, NumericArray]:
```
LeonidElkin, line 335: Validate that q_highest is greater than q_lowest

src/pysatl_core/distributions/strategies.py
```diff
@@ -168,66 +708,90 @@ def query_method(
...
+                        ),
+                    ),
+                    loop_method=fitted,
+                    edges=(),
+                )
+                injected_keys.append(intermediate_key)
+
+            # Remove the temporary loop plans injected for intermediate targets.
+            for key in injected_keys:
+                self._path_cache.pop(key, None)
```
LeonidElkin: I think it should be in the finally branch

LeonidElkin (changes requested):
The example is not the best, but in general is ok

LeonidElkin (approved):

Desiment (approved):

PR #77, opened 2026-03-05
Refactoring/computations

LeonidElkin:
- Cleaner work with analytical functions. They no longer require dummy parameters.
- Generics are now used only in abstractions that represent characteristics or computational methods.
- The `ComputationMethod` can now be used without a fictitious fit

### DIFF
--- a/src/pysatl_core/families/builtins/continuous/normal.py
+++ b/src/pysatl_core/families/builtins/continuous/normal.py
@@ -11,7 +11,7 @@ __copyright__ = "Copyright (c) 2025 PySATL project"
 __license__ = "SPDX-License-Identifier: MIT"
 
 import math
-from typing import TYPE_CHECKING, cast
+from typing import cast
 
 import numpy as np
 from scipy.special import erf, erfinv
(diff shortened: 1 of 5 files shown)


src/pysatl_core/families/builtins/continuous/normal.py
```diff
@@ -33,7 +33,7 @@
 )
 
 if TYPE_CHECKING:
-    from typing import Any
+    pass
```
Desiment: why we need such ``if`` here then?

src/pysatl_core/families/builtins/continuous/uniform.py
```diff
@@ -191,22 +191,22 @@ def char_func(parameters: Parametrization, t: NumericArray) -> ComplexArray:
...
         width = parameters.upper_bound - parameters.lower_bound
         return width**2 / 12
 
-    def skew_func(_1: Parametrization, _2: Any) -> int:
+    def skew_func(_1: Parametrization) -> int:
         """Skewness of uniform distribution (always 0)."""
         return 0
 
-    def kurt_func(_1: Parametrization, _2: Any, excess: bool = False) -> float:
+    def kurt_func(_1: Parametrization, *, excess: bool = False) -> float:
```
Desiment: rename ``_1`` -> ``parameters``?

PR #78, opened 2026-03-09
Transformations module implementation

LeonidElkin:
Base primitives and affine transformation for check. There is also an approximation, but it is very raw for now


Desiment (changes requested):

src/pysatl_core/transformations/operations/methods/affine.py
```diff
@@ -0,0 +1,347 @@
...
+) -> dict[GenericCharacteristicName, dict[LabelName, TransformationMethodSpec[Any, Any]]]:
+    """Merge transformation method dictionaries preserving per-target labels."""
+    merged: dict[
+        GenericCharacteristicName, dict[LabelName, TransformationMethodSpec[Any, Any]]
+    ] = {}
+    for method_set in method_sets:
+        for target, variants in method_set.items():
+            merged.setdefault(target, {}).update(variants)
+    return merged
+
```
Desiment: move to utils, remove duplicates

Desiment (changes requested):

Desiment (approved):

PR #84, opened 2026-04-05
Score function for builtins

MyznikovFD:
Add score (log‑PDF gradient) for built‑in distribution families

    Implement base_score for the base parametrization of Normal, Exponential, Uniform.

    Add gradient_transform to parametrization classes to propagate gradients via chain rule.

    Provide a unified score method in ParametricFamily that works for all parametrizations.

    Include parametrized tests verifying shape, type, and consistency across parametrizations.

### DIFF
--- a/src/pysatl_core/families/builtins/continuous/exponential.py
+++ b/src/pysatl_core/families/builtins/continuous/exponential.py
@@ -223,6 +223,38 @@ def configure_exponential_family() -> None:
         """Support of exponential distribution"""
         return ContinuousSupport(left=0.0)
 
+    def _base_score(parameters: Parametrization, x: NumericArray) -> NumericArray:
+        """
+        Compute the score (gradient of log‑PDF) for the base parametrization (rate λ).
+
+        For exponential distribution with rate λ > 0, log‑PDF is:
+            log f(x) = log λ - λ x   for x ≥ 0, else -∞.
+
+        The derivative with respect to λ is:
+            ∂/∂λ log f = 1/λ - x   (for x ≥ 0).
+
+        For points x < 0 the density is zero; we return 0 for numerical stability
+        (though the score is technically undefined there).
+
+        Parameters
+        ----------
+        parameters : Parametrization
+            Base parametrization instance (_Rate) with field lambda_.
+        x : NumericArray
+            Points at which to evaluate the gradient.
+
+        Returns
+        -------
+        NumericArray
+            Gradient array of shape (..., 1) where the last axis corresponds to
+            d(log f)/d(lambda_).
+        """
+        params = cast(_Rate, parameters)
+        lam = params.lambda_
+        inside = x >= 0
+        grad = np.where(inside, 1.0 / lam - x, 0.0)
+        return grad[..., np.newaxis]  # shape (..., 1)
+
     Exponential = ParametricFamily(
         name=FamilyName.EXPONENTIAL,
         distr_type=UnivariateContinuous,
@@ -239,6 +271,7 @@ def configure_exponential_family() -> None:
             CharacteristicName.KURT: kurt_func,
         },
         support_by_parametrization=_support,
+        base_score=_base_score,
     )
     Exponential.__doc__ = EXPONENTIAL_DOC
 
@@ -288,5 +321,28 @@ def configure_exponential_family() -> None:
                 Rate parametrization instance
             """
             return _Rate(lambda_=1.0 / self.beta)
+
+        def gradient_transform(self, grad_base: NumericArray) -> NumericArray:
+            """
+            Transform gradient from base parameter (rate λ) to scale β = 1/λ.
+
+            The transformation is: β = 1/λ.
+            Derivative: dβ/dλ = -1/λ².
+            Hence: grad_β = grad_λ * (-1/λ²) = -grad_λ / λ².
+
+            Parameters
+            ----------
+            grad_base : NumericArray
+                Gradient with respect to λ, shape (..., 1).
+
+            Returns
+            -------
+            NumericArray
+                Gradient with respect to β, shape (..., 1).
+            """
+            lam = 1.0 / self.beta  # λ = 1/β
+            grad_lam = grad_base[..., 0]
+            grad_beta = -grad_lam / (lam * lam)
+            return grad_beta[..., np.newaxis].astype(np.float64)
 
     ParametricFamilyRegister.register(Exponential)
--- a/src/pysatl_core/families/builtins/continuous/normal.py
+++ b/src/pysatl_core/families/builtins/continuous/normal.py
@@ -232,6 +232,36 @@ def configure_normal_family() -> None:
         """Support of normal distribution"""
         return ContinuousSupport()
 
+    def _base_score(parameters: Parametrization, x: NumericArray) -> NumericArray:
+        """
+        Compute the score (gradient of log‑PDF) for the base parametrization.
+
+        The base parametrization uses parameters mu (mean) and sigma (standard deviation).
+        The returned gradient has shape (..., 2) with the last axis corresponding
+        to [d(log f)/d(mu), d(log f)/d(sigma)].
+
+        Parameters
+        ----------
+        parameters : Parametrization
+            Base parametrization instance (must be _MeanStd).
+        x : NumericArray
+            Points at which the gradient is evaluated.
+
+        Returns
+        -------
+        NumericArray
+            Gradient array of shape (..., 2).
+        """
+        params = cast(_MeanStd, parameters)
+        mu = params.mu
+        sigma = params.sigma
+
+        z = (x - mu) / sigma
+
+        grad_mu = z / sigma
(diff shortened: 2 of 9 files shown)


src/pysatl_core/families/builtins/continuous/exponential.py
```diff
@@ -266,4 +348,27 @@ def transform_to_base_parametrization(self) -> Parametrization:
...
+
+            Returns
+            -------
+            NumericArray
+                Gradient with respect to β, shape (..., 1).
+            """
+            lam = 1.0 / self.beta  # λ = 1/β
+            grad_lam = grad_base[..., 0]
+            grad_beta = -grad_lam / (lam * lam)
+            return grad_beta[..., np.newaxis].astype(np.float64)
```
Desiment: This method is incompatible with idea of user defined parametrization; since it is assumes that any other parametrization from Rate will be always Scale

Desiment (changes requested):

src/pysatl_core/types.py
```diff
@@ -106,6 +106,9 @@ class EuclideanDistributionType(DistributionType):
 type NumericArray = NDArray[NumPyNumber]
 """Type alias for numeric arrays."""
 
+type NumericLike = NumericArray | Number
+"""Type alias for numeric array or scalar number."""
```
LeonidElkin: It doesn't really seem necessary to create such a type. We are already working with a system where only NumPy arrays are used. If you need to pass one element, then we pass an array of one element.

src/pysatl_core/families/builtins/continuous/normal.py
```diff
@@ -384,4 +432,99 @@ def transform_to_base_parametrization(self) -> Parametrization:
...
+        x_arr = np.atleast_1d(x)
+
+        if parameters.name == "meanStd":
+            return _base_score(parameters, x_arr)
+        base_params = parameters.transform_to_base_parametrization()
+        base_grad = _base_score(base_params, x_arr)
+        result = parameters.gradient_transform(base_grad)
+        return result
+
+    Normal.score = score  # type: ignore[method-assign]
```
LeonidElkin: It's not good to do that. We definitely don't want to dynamically add methods to objects. At least in this case. I think you need to optionally pass `score` to the family constructor, and then pull it in the `score` method of the class,, similar to how `support_resolver` is made. But don't name it `score_resolver`, just `score`

LeonidElkin (changes requested):

src/pysatl_core/families/parametric_family.py
```diff
@@ -418,4 +421,227 @@ def parametrization(
...
+            Points at which to evaluate the gradient.
+
+        Returns
+        -------
+        NumericArray
+            Gradient with respect to the parameters of the given parametrization.
+            Shape (..., d) where d is the number of parameters.
+        """
+        if self._score is None:
+            raise NotImplementedError(f"Family {self.name} does not implement score method")
```
LeonidElkin: I think it's probably more correct to raise ValueError instead of NotImplementedError

src/pysatl_core/families/parametric_family.py
```diff
@@ -418,4 +421,227 @@ def parametrization(
...
+        return self._score(parameters, x)
+
+    def view(
+        self,
+        parametrization_name: str | None = None,
+        **fixed: Any,
+    ) -> PartialParametricFamily:
```
LeonidElkin: Do not refer to it as `fixed`, but as `fixed_params` and make `parametrization_name` as a keyword-only argument via

```python
def view(
        self,
        *,
        parametrization_name: str | None = None,
        **fixed_params: Any,
    ) -> PartialParametricFamily:
```

src/pysatl_core/families/parametric_family.py
```diff
@@ -418,4 +421,227 @@ def parametrization(
...
+        return PartialParametricFamily(
+            family=self,
+            fixed_params=fixed,
+            parametrization_name=parametrization_name,
+        )
+
+    __call__ = distribution
+
+
+class PartialParametricFamily:
```
LeonidElkin: This class definitely must be an heir of the `ParametricFamily`

I will write all the problems in private messages because of their complexity

src/pysatl_core/families/parametric_family.py
```diff
@@ -418,4 +421,227 @@ def parametrization(
...
+        If any fixed parameter name is unknown for the specified parametrization,
+        or if the parametrization name is not registered in the family.
+    """
+
+    def __init__(
+        self,
+        family: ParametricFamily,
+        fixed_params: dict[str, Any],
+        parametrization_name: str | None = None,
+    ) -> None:
```
LeonidElkin: Rename `family` to `base_family`

src/pysatl_core/families/parametric_family.py
```diff
@@ -418,4 +421,227 @@ def parametrization(
...
+        """
+        target = parametrization_name or self._fixed_in_param
+        if target != self._fixed_in_param:
+            raise ValueError(
+                f"Fixed parameters are defined for parametrization '{self._fixed_in_param}'. "
+                f"Cannot create distribution in '{target}'. "
+                "Use the same parametrization or create a new view with "
+                "fix_params in that parametrization."
+            )
+        # Проверка конфликтов значений
```
LeonidElkin: Remember, No Russian

LeonidElkin (dismissed):

src/pysatl_core/families/builtins/continuous/exponential.py
```diff
@@ -223,6 +223,63 @@ def _support(_: Parametrization) -> ContinuousSupport:
...
+        Returns
+        -------
+        NumericArray
+            Gradient with respect to the parameters of the given parametrization.
+            Shape is (..., 1) for both rate and scale.
+        """
+        x_arr = np.atleast_1d(x)
+
+        if parameters.name == "rate":
+            return _base_score(parameters, x_arr)
```
Desiment: This if unnecessary

src/pysatl_core/families/builtins/continuous/exponential.py
```diff
@@ -223,6 +223,63 @@ def _support(_: Parametrization) -> ContinuousSupport:
...
+            Parametrization instance (rate or scale).
+        x : NumericArray
+            Points at which the gradient is evaluated.
+
+        Returns
+        -------
+        NumericArray
+            Gradient with respect to the parameters of the given parametrization.
+            Shape is (..., 1) for both rate and scale.
+        """
```
Desiment: This function can be moved into class of parametric family (assuming that we can differentiate with respect to all parameters which is true for now?)

src/pysatl_core/families/builtins/continuous/exponential.py
```diff
@@ -239,6 +296,7 @@ def _support(_: Parametrization) -> ContinuousSupport:
             CharacteristicName.KURT: kurt_func,
         },
         support_by_parametrization=_support,
+        score=score,
```
Desiment: we should pass ``_base_score`` instead, ``score`` functionality is duplicated

Desiment (changes requested):

tests/unit/families/builtins/continuous/test_normal.py
```diff
@@ -236,6 +236,122 @@ def test_normal_support(self):
...
+    def test_score_shape_and_type(self):
+        """Test that SCORE returns correct shape and type for all parametrizations."""
+        x = np.array([-1.0, 0.0, 1.0, 2.0, 3.0])
+
+        for param_name, params in [
+            ("meanStd", {"mu": 2.0, "sigma": 1.5}),
+            ("meanVar", {"mu": 2.0, "var": 2.25}),
+            ("meanPrec", {"mu": 2.0, "tau": 0.25}),
+            ("exponential", {"a": -0.222, "b": 0.888}),
+        ]:
```
LeonidElkin: Use instead parametric tests

LeonidElkin (changes requested):

tests/unit/families/builtins/continuous/test_normal.py
```diff
@@ -236,6 +236,85 @@ def test_normal_support(self):
...
+                    dmu_db := -1.0 / (2 * p["a"]),
+                    dsigma_da := 1.0 / (2 * np.sqrt(-2 * p["a"] ** 3)),
+                    dsigma_db := 0.0,
+                    grad_a := base_mu * dmu_da + base_sigma * dsigma_da,
+                    grad_b := base_mu * dmu_db + base_sigma * dsigma_db,
+                    np.stack([grad_a, grad_b], axis=-1),
+                )[-1],  # <-- берём последний элемент
+            ),
+        ],
+    )
```
LeonidElkin: Okay, this is too much. It's better to do multiple tests than such one.

LeonidElkin (changes requested):

LeonidElkin (approved):

Desiment (approved):

PR #85, opened 2026-04-06
Move DefaultSamplingUnivariateStrategy to sampling module

wrdxwrdxwrdx:
Moves the simple inverse-transform sampler out of distributions/strategies.py into its own sampling/simple.py module, where it logically belongs alongside other sampling implementations.


Desiment (changes requested):

LeonidElkin (approved):
LGTM!

Desiment (approved):

PR #86, opened 2026-04-23
View for fixed parameters

MyznikovFD:
Add `view` method for partial parameter fixing.

This PR introduces a `view` method that fixes a subset of parameters in any parametrization,
returning a `PartialParametricFamily` (a genuine `ParametricFamily` subclass).
The view exposes only the fixed parametrization as its local base, keeps all analytical
characteristics (via conversion to the original base when needed), and behaves like a
normal one‑parametrization family with fixed‑parameter metadata.

Example:
```python
norm = ParametricFamilyRegister.get("normal") 
norm_std1 = norm.view(parametrization_name="meanStd", std=1.0)          # fix std, mean is a variable
dist = norm_std1.distribution(mean=0.0) # Normal(0,1)

### DIFF
--- a/src/pysatl_core/families/parametric_family.py
+++ b/src/pysatl_core/families/parametric_family.py
@@ -15,12 +15,14 @@ __license__ = "SPDX-License-Identifier: MIT"
 import inspect
 from collections.abc import Mapping
 from functools import partial
-from typing import TYPE_CHECKING, Any, cast, dataclass_transform
+from types import MappingProxyType
+from typing import TYPE_CHECKING, Any, cast, dataclass_transform, overload
 
 import numpy as np
 
 from pysatl_core.distributions.computations.computation import AnalyticalComputation
 from pysatl_core.families.distribution import ParametricFamilyDistribution
+from pysatl_core.families.parametrizations import Parametrization
 from pysatl_core.types import (
     DEFAULT_ANALYTICAL_COMPUTATION_LABEL,
     ComputationFunc,
@@ -32,9 +34,6 @@ if TYPE_CHECKING:
 
     from pysatl_core.distributions.strategies import ComputationStrategy, SamplingStrategy
     from pysatl_core.distributions.support import Support
-    from pysatl_core.families.parametrizations import (
-        Parametrization,
-    )
     from pysatl_core.types import (
         GenericCharacteristicName,
         LabelName,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/pysatl_core/families/parametric_family.py [hunk 1]
```diff
@@ -451,4 +451,245 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+    ------
+    ValueError
+        If any fixed parameter name is unknown for the specified parametrization,
+        or if the parametrization name is not registered in the family.
+    """
+
+    _param_class: type[Parametrization]
+    _base_family: ParametricFamily
+    _fixed_params: dict[str, Any]
+    _fixed_in_param: str
```
LeonidElkin: it is not necessary to leave such notes if it isn't a dataclass. So you can remove it

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +451,245 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+        fixed_params: dict[str, Any],
+        parametrization_name: str | None = None,
+    ) -> None:
+        fixed_in = parametrization_name or base_family.base_parametrization_name
+
+        # Keep only the fixed parametrization in characteristics
+        filtered_chars = {}
+        for char_name, char_map in base_family.distr_characteristics.items():
+            if fixed_in in char_map:
+                filtered_chars[char_name] = {fixed_in: char_map[fixed_in]}
```
LeonidElkin: Can we automatically reduce characteristics that are not expressed through the fixed parameterization to the fixed one and fix the parameter for them already? After all, this is an analytically defined characteristic, it is strange that it disappears after fixing the parameters.

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +451,245 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+        Check whether all parameters of the fixed parametrization are already fixed.
+
+        Returns
+        -------
+        bool
+            True if every parameter of the parametrization has a fixed value,
+            False otherwise.
+        """
+        required_fields = set(getattr(self._param_class, "__dataclass_fields__", {}).keys())
+        return required_fields.issubset(self._fixed_params)
```
LeonidElkin: There should not be a situation where all the parameters are fixed. Then it should be `ParametricFamilyDistribution` object, not an object of this class.

There are two options: either you make a fallback in the `view` method to the `distribution` method, or you throw an exception when all the parameters are fixed and ask to use all the `distribution` method instead of the `view` method.

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +451,245 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+            name=base_family._name,
+            distr_type=base_family._distr_type,
+            distr_parametrizations=[fixed_in],
+            distr_characteristics=filtered_chars,
+            support_by_parametrization=base_family._support_resolver,
+            base_score=base_family._base_score,
+        )
+
+        param_class = base_family.get_parametrization(fixed_in)
+        self.register_parametrization(fixed_in, param_class)
```
LeonidElkin: This is incorrect logic, this object will have a different parameterization than the one that the parent had. Then we need to create a new parameterization class and specify it. You can leave the parent parameterization as some kind of `get_parent_parametrization()` method.

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +451,245 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+
+        self._fixed_in_param = fixed_in
```
LeonidElkin: Then why do we need the fixed_in intermediate variable?

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +451,245 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+        ------
+        KeyError
+            If `name` is not the fixed parametrization name.
+        """
+        if name != self._fixed_in_param:
+            raise KeyError(
+                f"Parametrization '{name}' is not available in this view. "
+                f"Only '{self._fixed_in_param}' is available."
+            )
+        return self._param_class
```
LeonidElkin: This is a normal approach to maintain compatibility with the base class, but we definitely need another method that simply returns this very single parameterization if the name is not passed. Well, that is, to make an overload
```suggestion
    @overload
    def get_parametrization(self) -> type[Parametrization]: ...

    @overload
    def get_parametrization(self, name: ParametrizationName) -> type[Parametrization]: ...

    def get_parametrization(
        self,
        name: ParametrizationName | None = None,
    ) -> type[Parametrization]:
        """
        Return the only parametrization class available in this partial family.

        If `name` is provided, it must match the fixed parametrization name.
        If `name` is omitted, the fixed parametrization class is returned.

        Parameters
        ----------
        name : ParametrizationName or None, optional
            Name of the parametrization. If omitted, the only available
            parametrization is returned.

        Returns
        -------
        type[Parametrization]
            The fixed parametrization class.

        Raises
        ------
        KeyError
            If `name` is provided and is not the fixed parametrization name.
        """
        if name is None:
            return self._param_class

        if name != self._fixed_in_param:
            raise KeyError(
                f"Parametrization '{name}' is not available in this view. "
                f"Only '{self._fixed_in_param}' is available."
            )

        return self._param_class
```

src/pysatl_core/families/parametric_family.py
```diff
@@ -418,4 +423,273 @@ def parametrization(
...
+        Raises
+        ------
+        ValueError
+            If not all parameters are fixed, or a conflict occurs.
+        """
+        if not self.is_complete():
+            required = set(getattr(self._param_class, "__dataclass_fields__", {}).keys())
+            missing = required - set(self._fixed_params)
+            raise ValueError(f"Not all parameters are fixed. Missing: {missing}")
+        return self.distribution(**kwargs)
```
LeonidElkin: Consider previous comments, must be deleted

src/pysatl_core/families/parametric_family.py
```diff
@@ -418,4 +423,273 @@ def parametrization(
...
+        additional_params : dict[str, Any]
+            Parameters to add or override.
+
+        Returns
+        -------
+        PartialParametricFamily
+            A new view with the combined fixed parameters.
+        """
+        new_fixed = {**self._fixed_params, **additional_params}
+        return PartialParametricFamily(self._base_family, new_fixed, self._fixed_in_param)
```
LeonidElkin, line 834: Can't we just reuse the parent methods? If not, then rename `with_fixed_params` to the `view`.

LeonidElkin (changes requested):
Add an example of using this functionality to the [overview](https://github.com/PySATL/pysatl-core/blob/main/examples/overview.ipynb) after the edits.

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +451,218 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+            )
+
+        super().__init__(
+            name=base_family._name,
+            distr_type=base_family._distr_type,
+            distr_parametrizations=[self._fixed_in_param],
+            distr_characteristics={},
+            support_by_parametrization=base_family._support_resolver,
+            base_score=base_family._base_score,
+        )
```
LeonidElkin, line 578: This view makes `self._fixed_in_param` the local base parametrization, but reuses
`base_family._distr_type` unchanged. The original callable is documented and used as accepting the
original family's base parametrization, so a view over a non-base parametrization can pass the
wrong parameter object. Please wrap it for the view, e.g. call
`base_family._distr_type(base_family.to_base(params))`, and add a test where `distr_type` depends
on a field that exists only on the original base parametrization.

```suggestion
```python
def _view_distr_type(params: Parametrization) -> DistributionType:
    canonical_params = base_family.to_base(params)
    return base_family._distr_type(canonical_params)

super().__init__(
    name=base_family._name,
    distr_type=_view_distr_type,
    distr_parametrizations=[self._fixed_in_param],
    distr_characteristics=...,
    support_by_parametrization=base_family._support_resolver,
    base_score=base_family._base_score,
)
```
```

File: src/pysatl_core/families/parametric_family.py [hunk 10]
```diff
@@ -451,4 +451,218 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+            ]
+            result[characteristic] = {
+                label_name: AnalyticalComputation(
+                    target=characteristic,
+                    func=self._bind_parametrization(func_factory, params_obj),
+                )
+                for label_name, func_factory in labeled_providers.items()
+            }
+
+        return result
```
LeonidElkin: Initializing the view with `distr_characteristics={}` makes this object inconsistent as a
`ParametricFamily`: `partial.distr_characteristics` and `partial._analytical_plan` say that no
analytical characteristics exist, while `_build_analytical_computations()` bypasses the view and
reads `_base_family`. Please build filtered view metadata instead: keep providers defined for
`self._fixed_in_param`, and wrap providers defined for the original base parametrization so they
are callable from the view's local base parameters.

You could do smth like this before `super().__init__`

```python
view_characteristics = {}
original_base = base_family.base_parametrization_name

for characteristic, forms in base_family.distr_characteristics.items():
    if self._fixed_in_param in forms:
        view_characteristics[characteristic] = {
            self._fixed_in_param: forms[self._fixed_in_param],
        }
    elif original_base in forms:
        view_characteristics[characteristic] = {
            self._fixed_in_param: _wrap_base_labeled_providers(forms[original_base]),
        }
```

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +451,218 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+        unknown = set(self._fixed_params) - required_fields
+        if unknown:
+            raise ValueError(
+                f"Unknown parameters for parametrization '{self._fixed_in_param}': {unknown}"
+            )
+
+    @property
+    def base(self) -> type[Parametrization]:
+        """Return the only parametrization class available in this view."""
+        return self._param_class
```
LeonidElkin: The view exposes the original parametrization class as `base`, but `distribution()` accepts a
different kwargs shape because fixed fields may be omitted. This weakens the "honest
ParametricFamily with one parametrization" model: `partial.base(**free_params)` cannot reproduce
the parameter object accepted by `partial.distribution(**free_params)`. Please either document that
`base` is the full concrete parametrization class, or introduce a wrapper/generated parametrization
class that represents only free parameters and injects fixed values when building the original full
parametrization.

for example 

```python
partial = fam.view(a=2.0)
partial.distribution(b=3.0)  # works fine
partial.base(b=3.0)          # failing: base parametrization of the view still need  parameter `a`
```

Generate a lightweight parametrization class view with only free fields. Its conversion logic
should substitute fixed values and then delegate the conversion logic to the initial
parameterization.

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +451,218 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+        *,
+        parametrization_name: str | None = None,
+        **additional_params: Any,
+    ) -> PartialParametricFamily:
+        if parametrization_name is not None and parametrization_name != self._fixed_in_param:
+            raise ValueError(
+                f"Cannot change parametrization. Current fixed parametrization is "
+                f"'{self._fixed_in_param}'. Use the same or omit the argument."
+            )
+        new_fixed = {**self._fixed_params, **additional_params}
```
LeonidElkin, line 833: `view()` on an existing partial view can overwrite already fixed parameters because
`additional_params` wins in `{**self._fixed_params, **additional_params}`. This should mirror the
conflict check in `distribution()` and raise if an already fixed parameter is passed with a
different value.


```suggestion
for key, fixed_val in self._fixed_params.items():
    if key in additional_params and additional_params[key] != fixed_val:
        raise ValueError(
            f"Parameter '{key}' is fixed to {fixed_val}, but got {additional_params[key]}"
        )

new_fixed = {**self._fixed_params, **additional_params}
```

tests/unit/families/test_partial_family.py
```diff
@@ -0,0 +1,201 @@
...
+        with pytest.raises(TypeError, match="unexpected keyword argument 'unknown'"):
+            partial.distribution(unknown=42)
+
+    # Analytical characteristics are preserved (indirect test)
+    def test_analytical_characteristics_preserved(self) -> None:
+        fam = self._make_two_param_family()
+        partial = fam.view(a=2.0)
+        dist = partial.distribution(b=3.0)
+        pdf_val = dist.calculate_characteristic(CharacteristicName.PDF, 0.0)
+        assert pdf_val == 5.0
```
LeonidElkin: This test covers analytical preservation only when the fixed parametrization is already the
original base. Please add a case for `view(parametrization_name="alt", ...)` where the
characteristic provider falls back to the original base parametrization. This should also cover the
currently untested `base_family.to_base(parameters)` branch and a regression test for callable
`distr_type`.

Suggested tests to add:

- A synthetic family with `base` and `alt` parametrizations where `PDF` is defined only for
  `base`.
- A partial view created with `view(parametrization_name="alt", ...)`.
- Distribution creation through that view.
- A characteristic calculation asserting that `alt -> base` conversion was used.
- A callable `distr_type` regression test where the callable reads a field that exists only on the
  original base parametrization.
- Smoke tests for built-in families, for example `Normal.meanVar` or
  `ContinuousUniform.meanWidth` partial views.

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +451,218 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+        self._fixed_in_param = parametrization_name or base_family.base_parametrization_name
+        param_class = base_family.get_parametrization(self._fixed_in_param)
+
+        # Check completeness: if all parameters are fixed, raise an error
+        required_fields = set(getattr(param_class, "__dataclass_fields__", {}).keys())
+        if required_fields.issubset(fixed_params):
+            raise ValueError(
+                f"All parameters of parametrization '{self._fixed_in_param}' are already fixed. "
+                "Use `.distribution()` directly to create a distribution instance."
+            )
```
LeonidElkin, line 547: Please validate unknown fixed parameter names before the "all parameters are fixed" check.
Otherwise a call that includes both all valid fields and an extra invalid field reports the less
useful completeness error and hides the unknown parameter.


```suggestion
required_fields = set(getattr(param_class, "__dataclass_fields__", {}).keys())

unknown = set(fixed_params) - required_fields
if unknown:
    raise ValueError(
        f"Unknown parameters for parametrization '{self._fixed_in_param}': {sorted(unknown)}"
    )

if required_fields.issubset(fixed_params):
    raise ValueError(
        f"All parameters of parametrization '{self._fixed_in_param}' are already fixed. "
        "Use `.distribution()` directly to create a distribution instance."
    )
```

LeonidElkin (changes requested):
Expected model for this feature:

- After fixing one or more parameters, the result should be a real `ParametricFamily` subclass.
- The resulting view should expose exactly one parametrization.
- The parametrization used for fixing parameters becomes the view's local base parametrization.
- The view may keep additional metadata about fixed parameters.
- Analytical characteristics should remain available whenever they can be expressed through the
  view parametrization.

In other words, the partial family should behave like a normal one-parametrization
`ParametricFamily`, with fixed-parameter metadata layered on top.

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +449,276 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+        return self._param_class
+
+    def _build_view_characteristics(self, base_family: ParametricFamily) -> dict[str, Any]:
+        view_chars = {}
+        original_base = base_family.base_parametrization_name
+
+        def wrap_provider(provider: Callable[..., Any]) -> Callable[..., Any]:
+            def wrapped(params: Parametrization, *args: Any, **kwargs: Any) -> Any:
+                base_params = base_family.to_base(params)
+                return provider(base_params, *args, **kwargs)
```
LeonidElkin: `_build_view_characteristics()` currently wraps original-base providers by always calling
`provider(base_params, *args, **kwargs)`. This bypasses the existing `_bind_parametrization()`
contract and breaks parametrization-independent providers such as `skew_func()` and keyword-only
providers such as `kurt_func(*, excess=False)`. Please bind the original provider with
`base_params` using the same `_bind_parametrization()` logic before invoking it.


```suggestion
def wrap_provider(
    provider: ParametricFamilyCharacteristic[Any, Any],
) -> ParametricFamilyCharacteristic[Any, Any]:
    def wrapped(params: Parametrization, *args: Any, **kwargs: Any) -> Any:
        base_params = base_family.to_base(params)
        bound = ParametricFamily._bind_parametrization(provider, base_params)
        return bound(*args, **kwargs)

    return wrapped
```

example of what would go wrong

```python
normal = registry.get(FamilyName.NORMAL)
partial = normal.view(parametrization_name="meanVar", mu=0.0)
dist = partial.distribution(var=1.0)
dist.query_method(CharacteristicName.SKEW)()
```

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +449,276 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+
+    def get_parametrization(self, name: ParametrizationName | None = None) -> type[Parametrization]:
+        if name is None:
+            return self._param_class
+        if name != self._fixed_in_param:
+            raise KeyError(
+                f"Parametrization '{name}' is not available in this view. "
+                f"Only '{self._fixed_in_param}' is available."
+            )
+        return self._param_class
```
LeonidElkin: The view now exposes a generated free-parameter class through `base`, but `parametrizations` and
`get_parametrization()` still return the original full parametrization class. The generated class
also has no fields `__param_name__`  and  `__family__`, so instances fail on `.name` and `to_base()`

- create `_free_param_class` before `super().__init__` or immediately after, but before registration;
- assign it `__param_name__ = self._fixed_in_param`;
- assign it `__family__ = self`;
- register it through `register_parametrization()`;
- make `parametrizations()` and `get_parametrization()` return
`_free_param_class`;

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +449,276 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+                raise ValueError(
+                    f"Parameter '{key}' is fixed to {fixed_val}, but got {kwargs[key]}"
+                )
+        all_params = {**self._fixed_params, **kwargs}
+        return super().distribution(
+            parametrization_name=target,
+            sampling_strategy=sampling_strategy,
+            computation_strategy=computation_strategy,
+            **all_params,
+        )
```
LeonidElkin, line 815: Same here `distribution()` should use `_free_param_class`. Now it still injects fixed values and calls the parent with an explicit `parametrization_name`, so the parent uses the original full parametrization class from `_parametrizations`

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +449,276 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+            f"{original_class.__name__}Free",
+            (Parametrization,),
+            {
+                "__init__": __init__,
+                "transform_to_base_parametrization": transform_to_base,
+                "__dataclass_fields__": {name: all_fields[name] for name in free_field_names},
+                "__annotations__": {name: all_fields[name].type for name in free_field_names},
+            },
+        )
+        return cast(type[Parametrization], new_class)
```
LeonidElkin: If the partial family exposes only free parameters, `score()` should return gradients with respect
to those free parameters only. The generated free-param class currently inherits the default
`gradient_transform()`, so score keeps the original/base dimensionality and includes fixed
coordinates. Adapt `gradient_transform()` for generated parametrizations

example of what wrong

```python
normal = registry.get(FamilyName.NORMAL)
partial = normal.view(parametrization_name="meanVar", mu=0.0)
free_params = partial.base(var=1.0)
partial.score(free_params, np.array([0.5]))
```

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +449,276 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+                    for label, provider in original_provider.items()
+                }
+                view_chars[char_name] = {self._fixed_in_param: wrapped}
+
+        return view_chars
+
+    def is_complete(self) -> bool:
+        """Check whether all parameters of the fixed parametrization are already fixed."""
+        required_fields = set(getattr(self._param_class, "__dataclass_fields__", {}).keys())
+        return required_fields.issubset(self._fixed_params)
```
LeonidElkin: Remove. There is no need in this after we make exceptions. It is just always true

src/pysatl_core/families/parametric_family.py
```diff
@@ -451,4 +449,276 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
...
+        return PartialParametricFamily(
+            base_family=self,
+            fixed_params=fixed_params,
+            parametrization_name=parametrization_name,
+        )
+
+    __call__ = distribution
+
+
+class PartialParametricFamily(ParametricFamily):
```
LeonidElkin, line 495: Add some properties containing probably useful metainformation about view and base family. My suggestions

```python
@property
def parent_family(self) -> ParametricFamily:
    """Original family this view was created from."""
    return self._base_family

@property
def fixed_parameters(self) -> Mapping[str, Any]:
    """Fixed parameter values."""
    return MappingProxyType(self._fixed_params)

@property
def fixed_parameter_names(self) -> frozenset[str]:
    """Names of fixed parameters."""
    return frozenset(self._fixed_params)

@property
def free_parameter_names(self) -> tuple[str, ...]:
    """Names of parameters that remain free in this view."""
    return self._free_parameter_names
```

LeonidElkin (changes requested):

LeonidElkin (approved):

Desiment (approved):

PR #107, opened 2026-05-10
Empirical Distribution

wrdxwrdxwrdx:
## Summary

Adds `EmpiricalDistribution` — a continuous univariate distribution fitted to observed
data through a pluggable density estimator (Gaussian KDE by default). The estimator
supplies both PDF and CDF analytically; PPF comes from a new tabulated `cdf -> ppf`
edge in the characteristic graph, registered globally rather than attached to this
distribution.

Usage: `examples/example_empirical_distribution.ipynb`.

## What's added

- **`EmpiricalDistribution`**, with `EmpiricalMethod` / `FittedEmpirical` protocols for
  custom estimators and a built-in `ScipyGaussianKde`. Exposes `data`, `method`,
  `estimator` and `tabulation_domain`; the estimator is swapped through `with_method`
  (independent clone) or `set_method` (in place). Samples are validated before fitting.
- **`cdf_to_ppf_tabulated_1C`** — a graph edge that tabulates the CDF once over a
  declared domain and inverts it by monotone interpolation, solving out-of-range
  quantiles exactly against the real CDF. Available to any distribution that declares a
  `tabulation_domain`.
- **`EmpiricalComputationStrategy`** — `DefaultComputationStrategy` with caching on by
  default and automatic invalidation when the estimator changes.
- **`DefaultUnuranSamplingStrategy.invalidate()` and `__deepcopy__`** — let the sampler
  survive method swaps and cloning.

## Non-obvious decisions

Each of these looks like it could be simplified. It can't.

**CDF is listed before PDF in `_build_analytical_computations`.** When a target has no
analytical self-loop, the resolver walks `analytical_computations` and takes the first
key from which *any* path exists — insertion order, not the shortest path. With PDF
first it picks `pdf -> cdf -> ppf` and rebuilds the CDF by quadrature even though an
exact one is right there: 987 ms versus 24 ms to fit the PPF, for identical results.
Pinned by `test_ppf_plan_starts_from_the_analytical_cdf`.

**The tabulated edge is registered before the general bisection edge.** `RegistryView`
keeps the first variant whose constraint admits the distribution, so ordering plus a
non-null `tabulation_domain` constraint is what routes empirical distributions to the
fast path and leaves everything else on bisection untouched.

**The fast PPF stays a graph edge rather than a method on the strategy.** Keeping it in
the graph is what makes `explain_computation_path` and `query_method` agree and lets
the standard option machinery reach the fitter.

**PCHIP, not a plain cubic spline.** A quantile function must be non-decreasing. On
smooth data either works, but on a CDF with a plateau a cubic spline overshoots and
produces a non-monotone PPF — a meaningless object. PCHIP guarantees monotonicity.

**Tails are solved exactly, the bulk is interpolated.** Where the tabulated CDF stops
changing, the grid points carry no usable information. Nudging them apart by an epsilon
to keep the interpolation going would invent data that is not there, so they are
dropped instead: the fast path covers the bulk of the distribution, and tail quantiles
fall through to the exact solver.

**A bounded `support` raises instead of being quietly ignored.** A Gaussian kernel has
no hard edges, so part of the fitted mass always lands outside a declared bound — a few
percent in practice. No estimator here can prevent that, and accepting the bound
anyway would hand back a distribution that assigns probability to impossible values.

**Cache invalidation is by estimator identity, not by an explicit call.** The strategy
compares the distribution's `estimator` by `is` before each query, so `set_method`
cannot leave a old fit behind through a forgotten `invalidate()`.

## Known gaps, not fixed here

- The leading tabulation knot is always kept, so some extreme left-tail quantiles get
  interpolated instead of falling through to the exact solver. Documented with a
  warning in the fitter.
- The resolver picks a source by insertion order rather than shortest path — see the
  first decision above.


src/pysatl_core/distributions/empirical/distribution.py
```diff
@@ -0,0 +1,567 @@
...
+        that into a ``ValueError`` naming the actual cause, and does so
+        uniformly for every :class:`EmpiricalMethod` — including ones that
+        would return unusable numbers rather than raise.
+
+        The check catches exact degeneracy only.  A sample whose spread is
+        merely tiny relative to its scale (identical values plus 1e-9 noise,
+        say) is degenerate in the same way but passes; guarding that needs a
+        scale-relative criterion and is left open deliberately.
+        """
+        sample = np.asarray(sample, dtype=float)
```
LeonidElkin: `np.asarray(sample, dtype=float)` retains the caller's buffer for a float64 array,
and `data` subsequently exposes that same writable array. SciPy's KDE also retains
the observations, and `_centers` is a view of its dataset. After a PPF has been
cached, modifying either `sample[:]` or `distr.data[:]` changes the live PDF/CDF
without replacing `estimator`, so the cached PPF, tabulation domain, and initialized
sampler still describe the old observations. For example, shifting the input array
by 100 after querying the median shifts the CDF but leaves the cached median
unchanged. This also affects every clone sharing the sample. Take an owned snapshot
before fitting and prevent writes through the public `data` accessor; clones can
share the protected snapshot. Add a regression covering mutations of both the
original input and the exposed data after PPF preparation.

src/pysatl_core/distributions/empirical/distribution.py
```diff
@@ -0,0 +1,567 @@
...
+        such references after calling :meth:`set_method`.
+
+        For side-by-side comparison of methods, prefer :meth:`with_method`.
+        """
+        self._method = method
+        self._estimator = method.fit(self._sample)
+        # Computation cache: auto-invalidated on next query via estimator-id
+        # tracking in EmpiricalComputationStrategy. Custom strategies that
+        # implement an invalidate() hook get a chance to reset, too.
+        getattr(self._computation_strategy, "invalidate", lambda: None)()
```
LeonidElkin: The optional `invalidate()` hook silently does nothing for
`DefaultComputationStrategy(enable_caching=True)`, which is accepted by the public
constructor but has no such method and no estimator-identity tracking. Querying
PPF with that strategy, calling `set_method()` with a different bandwidth, and
querying PPF again reuses the old fitted interpolator: the distribution and edge
identities in the cache key have not changed. Meanwhile the analytical PDF/CDF
already use the new estimator, so the characteristics disagree. Make cache
invalidation an implemented contract for the built-in caching strategies and invoke
it during the swap. Cover this sequence using an actual cached PPF from
`DefaultComputationStrategy`, rather than only a marker entry in the empirical
strategy's cache.

src/pysatl_core/distributions/computations/continuous.py
```diff
@@ -326,6 +346,245 @@ def _build_cdf_to_ppf_1C() -> FitterDescriptor:
...
+    # interpolation instead of being left to the exact solver.
+    cdf_grid = np.maximum.accumulate(cdf_grid)
+    usable = np.ones(cdf_grid.size, dtype=bool)
+    usable[1:] = np.diff(cdf_grid) > _MIN_TABULATED_CDF_STEP
+    q_knots = cdf_grid[usable]
+    x_knots = x_grid[usable]
+
+    if q_knots.size >= 2:
+        interpolator = _sp_interpolate.PchipInterpolator(q_knots, x_knots, extrapolate=False)
+        q_lowest, q_highest = float(q_knots[0]), float(q_knots[-1])
```
LeonidElkin: ### **This was found with AI help**

Dropping repeated CDF values does not make the inverse safe across a plateau:
PCHIP connects the point before the plateau to the first retained point after it,
and every probability between the global endpoints is still marked for
interpolation. Consider the continuous mixture
`0.5 * Uniform(0, 1) + 0.5 * Uniform(2, 3)`, with domain `(0, 3)` and the valid option
`grid_size=7`. The retained `(q, x)` knots are
`(0, 0), (0.25, 0.5), (0.5, 1), (0.75, 2.5), (1, 3)`.
The interpolated quantile at `q=0.55` lies between 1 and 2, where the CDF is exactly
0.5, although the correct quantile is 2.1. Increasing the grid size narrows the
affected probability interval but does not remove this failure. Split interpolation
into usable runs and send probabilities adjacent to filtered plateau runs to the
CDF solver. Add a disconnected-density regression checking `CDF(PPF(q))`, not just
monotonicity. This is an internal-plateau failure, separate from the documented
leading-knot tail limitation.

I guess @Desiment will agree or disagree with this finding

src/pysatl_core/distributions/empirical/distribution.py
```diff
@@ -0,0 +1,567 @@
...
+        Any external code that holds a direct reference to internal sampler
+        state (e.g. a value previously read from
+        ``distr.sampling_strategy._sampler``) keeps that reference alive and
+        will continue to sample from the *previous* distribution. Re-acquire
+        such references after calling :meth:`set_method`.
+
+        For side-by-side comparison of methods, prefer :meth:`with_method`.
+        """
+        self._method = method
+        self._estimator = method.fit(self._sample)
```
LeonidElkin: `self._method` is changed before `method.fit()` can raise. If fitting fails, the
object reports the rejected method through its public `method` property while
retaining the previous estimator, cached computations, and sampler. A caller that
catches a custom estimator's fitting error therefore observes inconsistent state,
and a subsequent strategy-only clone copies that mismatched method/estimator pair.
Fit into a local variable first, then publish the method and fitted estimator
together before invalidating derived state. Add a failing-method test asserting
that both public properties and the original characteristics remain unchanged.

src/pysatl_core/distributions/empirical/distribution.py
```diff
@@ -0,0 +1,567 @@
...
+        self._weights = np.asarray(kde.weights, dtype=float)
+
+    def pdf(self, x: NDArray[np.float64]) -> NDArray[np.float64]:
+        scalar_input = np.ndim(x) == 0
+        x_arr = np.atleast_1d(np.asarray(x, dtype=float))
+        finite = np.isfinite(x_arr)
+        result = np.zeros_like(x_arr)
+        if finite.any():
+            result[finite] = self._kde.pdf(x_arr[finite])
+        return result[0] if scalar_input else result
```
LeonidElkin: The finite mask excludes `NaN` as well as infinities, but the result is initialized
to zero. Consequently `pdf(np.array([np.nan]))` returns zero, while this estimator's
CDF propagates NaN. Zero is a legitimate density value and silently disguises
invalid query data, potentially turning an upstream numerical failure into an
apparently valid likelihood. Keep the zero result for positive and negative
infinity, and explicitly propagate NaN at NaN positions. Cover a mixed input of
finite values, both infinities, and NaN.

tests/unit/distributions/empirical/test_strategy.py
```diff
@@ -0,0 +1,418 @@
...
+        sample = rng.normal(0.0, 1.0, 400)
+
+        fast = EmpiricalDistribution(sample)
+        slow = EmpiricalDistribution(sample, computation_strategy=DefaultComputationStrategy())
+        quantiles = np.linspace(0.1, 0.9, 9)
+
+        fast_vals = fast.calculate_characteristic(CharacteristicName.PPF, quantiles)
+        slow_vals = np.array(
+            [
+                float(np.squeeze(slow.calculate_characteristic(CharacteristicName.PPF, float(q))))
```
LeonidElkin: **This was found with AI help**

`slow` still exposes `tabulation_domain`, and the tabulated edge is selected by the
global registry constraint, not by `EmpiricalComputationStrategy`. Both objects in
this test therefore execute `cdf_to_ppf_tabulated_1C`; changing the strategy only
changes caching. The test named `test_matches_root_finding_within_tolerance` compares
the same interpolator evaluated in vector and scalar calls, so a common quantile
error passes unnoticed. Build the reference with an independent root solver against
the analytical CDF and compare its quantiles with the tabulated results. Include
multimodal data and probabilities near plateaus to exercise the approximation's
failure modes.

src/pysatl_core/sampling/unuran/core/unuran_sampling_strategy.py
```diff
@@ -94,6 +94,53 @@ def sample(self, n: int, distr: Distribution, **options: Any) -> NumericArray:
...
+        Notes
+        -----
+        The configuration object is shared, not copied.  It is a frozen
+        dataclass, so its fields cannot be rebound, but ``method_params`` is a
+        plain dict and mutating it in place is visible from both strategies.
+        Nothing in the library mutates it today.
+        """
+        new = DefaultUnuranSamplingStrategy(config=self._config_value)
+        memo[id(self)] = new
+        return new
```
LeonidElkin: The new `__deepcopy__` correctly drops the native sampler but reuses
`self._config_value`. `UnuranMethodConfig` is only shallowly frozen:
`method_params` is a mutable dictionary and may contain mutable values. Mutating
`clone.sampling_strategy.config.method_params` therefore also changes the original
strategy's configuration, despite the distribution cloning API promising
independent strategies. This changes the behavior of deep-copying an uninitialized
strategy as well, where the former default implementation copied the configuration.
Deep-copy the configuration using the supplied memo while continuing to discard
the native sampler. Add a clone-isolation check with a nested parameter value.

LeonidElkin (changes requested):

PR #118, opened 2026-05-25
fix(families): raise ValueError for x outside support in builtins

MyznikovFD:


src/pysatl_core/families/builtins/continuous/uniform.py
```diff
@@ -291,11 +290,12 @@ def _base_score(parameters: Parametrization, x: NumericArray) -> NumericArray:
...
-        width = b - a
+        if np.any((x < a) | (x > b)):
+            raise ValueError(f"Score is undefined for x outside support [{a}, {b}]. Got x = {x}")
 
-        inside = (x >= a) & (x <= b)
-        grad_a = np.where(inside, 1.0 / width, 0.0)
-        grad_b = np.where(inside, -1.0 / width, 0.0)
+        width = b - a
+        grad_a = np.full_like(x, 1.0 / width)
+        grad_b = np.full_like(x, -1.0 / width)
```
LeonidElkin: `np.full_like` preserves the dtype of `x`. For integer inputs, `1.0 / width` and `-1.0 / width` are cast to integers, so the score becomes `[[0, 0], ...]` instead of the expected floating-point gradient. Force a floating dtype here, e.g. `np.full_like(x, ..., dtype=np.float64)`, and add a regression test with integer `x`.

src/pysatl_core/families/builtins/continuous/exponential.py
```diff
@@ -251,9 +250,10 @@ def _base_score(parameters: Parametrization, x: NumericArray) -> NumericArray:
         """
         params = cast(_Rate, parameters)
         lam = params.lambda_
-        inside = x >= 0
-        grad = np.where(inside, 1.0 / lam - x, 0.0)
-        return grad[..., np.newaxis]  # shape (..., 1)
+        if np.any(x < 0):
+            raise ValueError(f"Score is undefined for x < 0 (outside support). Got x = {x}")
```
LeonidElkin: This check only rejects `x < 0`, but the family support excludes non-finite values as well. For example, `ContinuousSupport(left=0.0).contains(np.inf)` is false, while `score([np.inf])` currently returns `-inf` instead of raising. Reject non-finite `x` values, ideally with a test for `np.inf` and `np.nan`.

src/pysatl_core/families/builtins/continuous/uniform.py
```diff
@@ -291,11 +290,12 @@ def _base_score(parameters: Parametrization, x: NumericArray) -> NumericArray:
         params = cast(_Standard, parameters)
         a = params.lower_bound
         b = params.upper_bound
-        width = b - a
+        if np.any((x < a) | (x > b)):
+            raise ValueError(f"Score is undefined for x outside support [{a}, {b}]. Got x = {x}")
```
LeonidElkin, line 304: This support check does not reject `NaN`: both `(x < a)` and `(x > b)` are false for `NaN`, so `score([np.nan])` returns a regular-looking gradient. Since `NaN` is not contained in the distribution support, reject non-finite values before computing the gradient.

src/pysatl_core/families/parametric_family.py
```diff
@@ -438,6 +438,9 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
         NumericArray
             Gradient with respect to the parameters of the given parametrization.
             Shape is (..., d), where d is the number of parameters of the parametrization.
+
+        ValueError
+            If any value in `x` lies outside the distribution's support.
```
LeonidElkin: This should be documented under a NumPy-style `Raises` section. Right now `ValueError` is placed directly after `Returns`, so the docstring structure is invalid and may render incorrectly in generated docs.

LeonidElkin (commented):
The issue explicitly lists `normal.py`, but this commit does not add any support validation for the normal score. Even though all finite real values are in support, `ContinuousSupport()` excludes `±inf` and `NaN`, and the current implementation returns `inf`/`nan` gradients for those inputs. Either add a finite-value check here or clarify why normal is intentionally out of scope.

src/pysatl_core/families/parametric_family.py
```diff
@@ -438,6 +438,11 @@ def score(self, parameters: Parametrization, x: NumericArray) -> NumericArray:
         NumericArray
             Gradient with respect to the parameters of the given parametrization.
             Shape is (..., d), where d is the number of parameters of the parametrization.
+
+        Raises
+        ------
+        ValueError
+            If any value in `x` lies outside the distribution's support.
```
LeonidElkin: The generic `score()` docstring now documents support validation as if it were enforced by
`ParametricFamily.score()` itself, but this method only delegates to `_base_score`. Custom families can
still return any result for out-of-support `x`, and this method also raises `ValueError` when the family
does not provide a score implementation at all. Keep the public contract accurate and document the
delegated failure mode instead of promising support checks for every family.

```suggestion
        Raises
        ------
        ValueError
            If this family does not provide a score implementation. Family-specific
            score implementations may also raise ValueError when the score is undefined
            for the supplied evaluation points, such as points outside support.
```

LeonidElkin (dismissed):

LeonidElkin (approved):

PR #121, opened 2026-07-16
Families/gamma family

### LINKED ISSUE #41: Family: Gamma
# Family: Gamma

Naming: `Gamma` (gamma distribution).

## Task

Implement the gamma family as a `ParametricFamily`, add all parameterizations and analytical characteristics, and register the family in the registry.

## Base parameterization

- Base parameters: `(k, θ)` with `k > 0` (shape), `θ > 0` (scale).

## Other parameterizations and relations

1. Shape–rate `(k, β)`:
   - Parameters: `k > 0`, `β > 0` (rate).
   - Relations to base:
     - `θ = 1 / β`
     - `β = 1 / θ`.

2. Mean–variance `(m, v)`:
   - Parameters: `m > 0` (mean), `v > 0` (variance).
   - Relations to base:
     - `k = m^2 / v`
     - `θ = v / m`.
   - Inverse relations:
     - `m = k * θ`
     - `v = k * θ^2`.

All characteristics should be defined in terms of `(k, θ)`.

## Analytical characteristics (in terms of k, θ)

For `x > 0`:

- `pdf(x)`:
  - `pdf(x) = x^(k - 1) * exp(-x / θ) / (Γ(k) * θ^k)`.

- `cdf(x)`:
  - `cdf(x) = γ(k, x / θ) / Γ(k)`,  
    where `γ` is the lower incomplete gamma function.

- `ppf(u)` (0 < u < 1):
  - `ppf(u)` is the inverse function of `cdf(x)` with respect to `x` for fixed `(k, θ)`.

- `cf(t)`:
  - `cf(t) = (1 - i * θ * t)^(-k)`.

- `lpdf(x)`:
  - `lpdf(x) = (k-1)*log(x) - x/θ - k*log(θ) - lgamma(k) 

- `mean`:
  - `mean = k * θ`.

- `var`:
  - `var = k * θ^2`.

BykovaElizaveta:
- Implement Gamma distribution with shapeScale, shapeRate, meanVar parametrizations

- Add analytical implementations for PDF, CDF, PPF, characteristic function, log-PDF
 
- Implement moment calculations (mean, variance)
 
- Configure family registry, parametrization system, and continuous support (left=0.0)
 
- Add tests covering all distribution characteristics
 
- Test all three parametrizations and conversions between them
 
- Verify against scipy.stats.gamma for correctness

Resolves https://github.com/PySATL/pysatl-core/issues/41

### DIFF
--- a/src/pysatl_core/families/builtins/continuous/gamma.py
+++ b/src/pysatl_core/families/builtins/continuous/gamma.py
@@ -1,0 +1,400 @@
+from __future__ import annotations
+
+__author__ = "Elizaveta Bykova"
+__copyright__ = "Copyright (c) 2025 PySATL project"
+__license__ = "SPDX-License-Identifier: MIT"
+
+from typing import cast
+
+import numpy as np
+from scipy.special import digamma, gammainc, gammaincinv, gammaln
+
+from pysatl_core.distributions.support import ContinuousSupport
+from pysatl_core.families.parametric_family import ParametricFamily
+from pysatl_core.families.parametrizations import (
+    Parametrization,
+    constraint,
+    parametrization,
+)
+from pysatl_core.families.registry import ParametricFamilyRegister
+from pysatl_core.types import (
+    CharacteristicName,
+    ComplexArray,
+    FamilyName,
+    NumericArray,
+    UnivariateContinuous,
+)
+
+
+def configure_gamma_family() -> None:
+    """
+    Configure and register the Gamma distribution family.
+    """
+
+    if ParametricFamilyRegister.contains(FamilyName.GAMMA):
+        return
+
+    GAMMA_DOC = """
+    Gamma distribution.
+
+    The gamma distribution is a two-parameter family of continuous probability
+    distributions defined on the non-negative real line [0, +inf). It is
+    asymmetric and is characterized by two parameters:
+    shape (k) and scale (θ), both of which must be positive.
+
+    Probability density function:
+        f(x) = x^(k-1) * exp(-x/θ) / (Γ(k) * θ^k)  for x > 0
+
+    The gamma distribution is widely used in statistics, engineering, and science to model
+    waiting times, life testing, and continuous variables that are non-negative and
+    right-skewed.
+    """
+
+    def pdf(parameters: Parametrization, x: NumericArray) -> NumericArray:
+        """
+        Probability density function for gamma distribution.
+
+        Parameters
+        ----------
+        parameters : Parametrization ()
+        Distribution parameters object with fields:
+            - k: float (shape)
+            - theta: float (scale)
+        x : NumericArray
+            Points at which to evaluate the probability density function
+
+        Returns
+        -------
+        NumericArray
+            Probability density values at points x
+        """
+        return cast(NumericArray, np.exp(lpdf(parameters, x)))
+
+    def cdf(parameters: Parametrization, x: NumericArray) -> NumericArray:
+        """
+        Cumulative distribution function for gamma distribution.
+
+        Parameters
+        ----------
+        parameters : Parametrization
+            Distribution parameters object with fields:
+            - k: float (shape)
+            - theta: float (scale)
+        x : NumericArray
+            Points at which to evaluate the cumulative distribution function
+
+        Returns
+        -------
+        NumericArray
+            Probabilities P(X ≤ x) for each point x
+        """
+        parameters = cast(_ShapeScale, parameters)
+        k = parameters.k
+        theta = parameters.theta
+
+        x_arr = np.asarray(x, dtype=np.float64)
+        safe_x = np.maximum(x_arr, 0.0)
+        return cast(NumericArray, np.where(x_arr > 0, gammainc(k, safe_x / theta), 0.0))
+
+    def ppf(parameters: Parametrization, u: NumericArray) -> NumericArray:
+        """
+        Percent point function (inverse CDF) for gamma distribution.
+
+        Parameters
+        ----------
+        parameters : Parametrization
+            Distribution parameters object with fields:
+            - k: float (shape)
+            - theta: float (scale)
+        u : NumericArray
+            Probability from [0, 1]
+
+        Returns
+        -------
+        NumericArray
+            Quantiles corresponding to probabilities p
+            If p[i] is 0 or 1, then the result[i] is 0.0 and inf correspondingly
+
+        Raises
+        ------
+        ValueError
+            If probability is outside [0, 1]
+        """
+        if np.any((u < 0) | (u > 1)):
+            raise ValueError("Probability must be in [0, 1]")
+        parameters = cast(_ShapeScale, parameters)
+
+        k = parameters.k
+        theta = parameters.theta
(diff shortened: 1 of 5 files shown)


src/pysatl_core/families/builtins/continuous/gamma.py
```diff
@@ -0,0 +1,315 @@
...
+        distr_characteristics={
+            CharacteristicName.PDF: pdf,
+            CharacteristicName.CDF: cdf,
+            CharacteristicName.PPF: ppf,
+            CharacteristicName.CF: char_func,
+            CharacteristicName.LPDF: lpdf,
+            CharacteristicName.MEAN: mean_func,
+            CharacteristicName.VAR: var_func,
+        },
+        support_by_parametrization=_support,
```
MyznikovFD, line 261: `base_score` should be provided in the family creation. I wrote more detailed instruction in the issue #41

BykovaElizaveta: done

MyznikovFD: so you can resolve the conversation:)

src/pysatl_core/families/builtins/continuous/gamma.py
```diff
@@ -0,0 +1,385 @@
...
+        Raises
+        ------
+        ValueError
+            If any x <= 0 (outside support).
+        """
+        parameters = cast(_ShapeScale, parameters)
+        x_arr = np.asarray(x, dtype=np.float64)
+
+        if np.any(x_arr <= 0):
+            raise ValueError("base_score is undefined for x <= 0 (outside gamma support)")
```
MyznikovFD: This check only rejects x < = 0, but the family support excludes non-finite values as well. For example, ContinuousSupport(left=0.0).contains(np.inf) is false, while score([np.inf]) currently returns -inf instead of raising. Reject non-finite x values, ideally with a test for np.inf and np.nan
I made the similar mistake in https://github.com/PySATL/pysatl-core/pull/118, so you can check how I resolved the problem. My PR is not merged in main, though, so you may find some other way.

src/pysatl_core/families/builtins/continuous/gamma.py
```diff
@@ -0,0 +1,385 @@
...
+            Log-probability density values at points x
+        """
+        parameters = cast(_ShapeScale, parameters)
+
+        k = parameters.k
+        theta = parameters.theta
+
+        result = np.full_like(x, -np.inf, dtype=np.float64)
+        mask = x > 0
+        x_pos = x[mask]
```
LeonidElkin: `lpdf()` indexes the original `x` value instead of a NumPy array. A scalar call such as
`dist.query_method(CharacteristicName.LPDF)(1.0)` passes a Python float through the bound analytical
method, so `x[mask]` raises instead of returning the scalar log-density. The same mask also treats
`np.inf` as a positive point, and line 180 then evaluates an `inf - inf` expression, producing `nan`;
`pdf()` inherits that `nan` because it is implemented as `np.exp(lpdf(...))`. Normalize the input once
and compute the density only for finite positive values.

```suggestion
        x_arr = np.asarray(x, dtype=np.float64)
        result = np.full_like(x_arr, -np.inf, dtype=np.float64)
        mask = np.isfinite(x_arr) & (x_arr > 0)
        x_pos = x_arr[mask]
```

src/pysatl_core/families/builtins/continuous/gamma.py
```diff
@@ -0,0 +1,385 @@
...
+        Returns
+        -------
+        NumericArray
+            Gradient array of shape (..., 2) with partial derivatives
+            [d lpdf/dk, d lpdf/dtheta].
+
+        Raises
+        ------
+        ValueError
+            If any x <= 0 (outside support).
```
LeonidElkin: The distribution support is `[0, +inf)`, so `x = 0` is a boundary point, not outside support. The score
formula is still undefined there because it contains `log(x)`, but the docstring and error text should
not describe the boundary as outside gamma support. Use wording that separates the distribution support from the score domain. This is exactly how it is done in SciPy and described on Wikipedia. This was not originally stated in the issue. I edited it.

```suggestion
            If any x <= 0 or any x is non-finite. The score is undefined at
            the x = 0 boundary and outside support.
```

src/pysatl_core/families/builtins/continuous/gamma.py
```diff
@@ -0,0 +1,391 @@
...
+    defined on the positive real line. It is asymmetric and is characterized by two parameters:
+    shape (k) and scale (θ), both of which must be positive.
+
+    Probability density function:
+        f(x) = x^(k-1) * exp(-x/θ) / (Γ(k) * θ^k)  for x > 0
+
+    The gamma distribution is widely used in statistics, engineering, and science to model
+    waiting times, life testing, and continuous variables that are strictly positive and
+    right-skewed.
+    """
```
LeonidElkin, line 51: The docs still describe Gamma as defined on the "positive real line", and the support
test docstring says `(0, inf)`, while the agreed support is `[0, +inf)`. The code asserts
`left_closed`, so the implementation and the documentation are now out of sync.

Update these descriptions to say non-negative support / `[0, +inf)`, and keep the
`x > 0` wording only for formulas that contain `log(x)`.

src/pysatl_core/families/builtins/continuous/gamma.py
```diff
@@ -0,0 +1,391 @@
...
+        k = parameters.k
+        theta = parameters.theta
+
+        x_arr = np.asarray(x, dtype=np.float64)
+        result = np.full_like(x_arr, -np.inf, dtype=np.float64)
+        mask = np.isfinite(x_arr) & (x_arr > 0)
+        x_pos = x_arr[mask]
+
+        result[mask] = (k - 1) * np.log(x_pos) - x_pos / theta - k * np.log(theta) - gammaln(k)
+        return cast(NumericArray, result)
```
LeonidElkin, line 192: `lpdf()` initializes every non-positive input to `-inf` and only computes values for
finite `x > 0`. Because `pdf()` is implemented as `np.exp(lpdf(...))`, both functions
return `0` at `x = 0` for every shape parameter.

That contradicts the updated Gamma issue contract: for `0 < k < 1`, `lpdf(0)` and
`pdf(0)` must be `+inf`; for `k = 1`, `lpdf(0)` must be `-log(theta)` and `pdf(0)`
must be `1 / theta`; only for `k > 1` should they be `-inf` and `0`. The current tests
only cover the `k > 1` branch, so the broken `0 < k <= 1` behavior would pass.

Handle the `x == 0` boundary explicitly in `lpdf()` before returning, and add boundary
tests for `k < 1`, `k == 1`, and `k > 1`.

LeonidElkin (changes requested):

src/pysatl_core/families/builtins/continuous/gamma.py
BykovaElizaveta: @LeonidElkin Thank you for highlighting the boundary asymptotics

LeonidElkin (approved):
Good job!

